PR #166, opened 2026-05-15
fix(plot): evidence-gaps provenance + intervention-target exclusion

Talchain:
## Summary

- **m1_coaching.evidence_gaps** now consumes the same enriched, PLoT-recomputed signals that the public `factor_sensitivity[]` already publishes — no more raw-ISL confidence/influence republished under public coaching field names.
- **Intervention-target factors** (decision levers) are derived from `options[i].interventions` keys and excluded from the evidence-gaps list **before** the rank/quartile gate, so they cannot crowd out genuine background-uncertainty factors that ranked below them in the enriched importance order.
- **No silent backfill:** if no non-lever factor passes the existing confidence/VOI gates, evidence_gaps is honestly empty.

## What changed (files)

| File | Change |
|---|---|
| [src/coaching/types.ts](https://github.com/Talchain/plot-lite-service/blob/claude-plot/evidence-gap-provenance/src/coaching/types.ts) | New optional `CoachingInputs.interventionTargetIds: Set<string>` |
| [src/coaching/normalise-inputs.ts](https://github.com/Talchain/plot-lite-service/blob/claude-plot/evidence-gap-provenance/src/coaching/normalise-inputs.ts) | New optional 4th param `enrichedFactorSensitivity`; maps from enriched array when supplied. Derives `interventionTargetIds` from union of `options[i].interventions` keys. |
| [src/coaching/m1-coaching.ts](https://github.com/Talchain/plot-lite-service/blob/claude-plot/evidence-gap-provenance/src/coaching/m1-coaching.ts) | New optional 8th param `factorSensitivityEnriched: FactorSensitivityResultV3[]`; threaded through to `normaliseCoachingInputs`. |
| [src/coaching/evidence-gaps.ts](https://github.com/Talchain/plot-lite-service/blob/claude-plot/evidence-gap-provenance/src/coaching/evidence-gaps.ts) | Exclude levers **before** computing `k` and applying the rank/quartile gate (operator-precedence fix). |
| [src/routes/v2/run.ts](https://github.com/Talchain/plot-lite-service/blob/claude-plot/evidence-gap-provenance/src/routes/v2/run.ts) | Pass the already-built merged `factorSensitivity` variable to `generateM1Coaching` so coaching sees identical provenance to the public payload. |
| [tests/coaching/evidence-gaps-provenance.test.ts](https://github.com/Talchain/plot-lite-service/blob/claude-plot/evidence-gap-provenance/tests/coaching/evidence-gaps-provenance.test.ts) | New file: 17 tests covering provenance, derivation, composition order, no-backfill, legacy fallback. |

Diff: +487 / -22 across 6 files. No schema changes, no new public fields, no OpenAPI changes.

## Provenance explained

The audit truth-table row A1-PRIMARY requires that the public `factor_sensitivity[].confidence` value be PLoT-recomputed (`0.5 × band_score(attribution_stability) + 0.5 × mean(exists_probability)`), not ISL's own `confidence` field. The merge step at [src/lib/factor-influence.ts:761](https://github.com/Talchain/plot-lite-service/blob/staging/src/lib/factor-influence.ts#L761) explicitly strips ISL's confidence before emission.

Pre-this-PR, `m1_coaching.evidence_gaps` bypassed that merge entirely — it called `normaliseCoachingInputs` with raw ISL `factor_sensitivity`, which preserved raw-ISL `confidence` and raw-ISL `elasticity` and republished them as `evidence_gaps[].confidence` and `evidence_gaps[].influence` in the public response. Two different `confidence` values shipped under the same field name in the same response.

This PR feeds the same merged array (the public payload's source) into coaching, so the two surfaces are byte-identical in their provenance.

## Intervention-target source-of-truth

The lever set is derived from `options[i].interventions` — the request-side intervention map. Raw ISL `zero_reason='intervention_override'` is **not** the source of truth; it is a downstream symptom that can mis-fire when ISL's elasticity estimator drifts. A unit test pins this invariant ([test: 'raw ISL zero_reason is NOT required for the exclusion to work'](https://github.com/Talchain/plot-lite-service/blob/claude-plot/evidence-gap-provenance/tests/coaching/evidence-gaps-provenance.test.ts)).

## Composition order (operator precedence)

The legacy filter chain was:
```
top-quartile-by-importance ∩ confidence<0.7 ∩ voi≥floor ∩ NOT lever
```
applied with `top-quartile` first. When levers consume the top ranks (as they often do in the graph-based importance order), the rank gate leaves "holes" that legitimate non-lever factors below them cannot fill. This was the regression caught in the stop-and-report iteration — the assistant bundle's only non-lever (`fac_strategic_value`) was at rank 4 of 4, and the rank gate `rank ≤ 3` excluded it.

Fix: apply `NOT lever` first, then compute `k` and the rank gate over the non-lever subset. Same predicates, correct operator order.

## Before/after evidence (saved bundles, local replay, no staging hit)

**Bundle 1 — staging-assistant `c1abcea5`** (3 of 4 factors are levers from `opt_ai_only`, `opt_human_ai`, etc.):

| Rank | BEFORE (production) | AFTER (this PR) |
|---|---|---|
| 1 | `fac_strategic_value` voi=0.6351 conf=0.3649 inf=1.0 | `fac_strategic_value` voi=0.1042 **conf=0.375** **inf=0.167** |
| 2 | `fac_annual_cost` voi=0 conf=0.10 inf=0 *(filler)* | — *(raw-ISL filler entry dropped)* |
| 3 | `fac_onboarding_time` voi=0 conf=0.10 inf=0 *(filler)* | — *(raw-ISL filler entry dropped)* |

Same top factor, with PLoT-recomputed confidence/influence. The two zero-VOI raw-ISL "filler" entries that ranked into evidence_gaps under the legacy path are correctly absent (they only passed under raw-ISL because raw-ISL importance_rank put them in the top quartile).

**Bundle 2 — dev-hiring/expansion `c952d8d6`** (3 of 5 factors are levers):

| Rank | BEFORE (production) | AFTER (this PR) |
|---|---|---|
| 1 | `fac_competition` voi=0.6173 conf=0.3827 inf=1.0 | `fac_competition` voi=0.5686 **conf=0.375** **inf=0.9097** |
| 2 | `fac_current_customers` voi=0.2044 conf=0.3659 inf=0.32 | `fac_current_customers` voi=0.1522 **conf=0.375** **inf=0.2436** |
| 3 | `fac_expansion_investment` voi=0 conf=0.10 inf=0 *(filler)* | — *(raw-ISL filler entry dropped)* |

Same top two factors in the same order. PLoT-recomputed confidence/influence. Zero-VOI lever filler dropped.

**Lever-exclusion check:** zero intervention-target factors appear in either AFTER list. **No silent backfill.**

## Invariants preserved (audit checklist)

| Invariant | Status |
|---|---|
| Public `factor_sensitivity[]` array contents (graph-merged, recomputed confidence) | unchanged |
| `factor_sensitivity[].confidence_provenance` audit metadata | unchanged |
| B3 `_meta.auto_noise_applied` disclosure | unchanged |
| `evpi_percentage_points` non-negativity (Howard 1966) | unchanged |
| OpenAPI schema | unchanged (no new fields) |
| ISL request / CEE request shape | unchanged |
| Pre-existing 4683 tests | pass |

The only public-payload field changed by this PR is **`m1_coaching.evidence_gaps`**.

## Test results

- `npx tsc --noEmit` — clean.
- `npx vitest run tests/coaching/` — 75/75 pass (58 existing m1-coaching + 17 new provenance).
- `bash scripts/pre-push-validate.sh` — **4683 passed, 25 skipped, 0 failed**. Branch guard, typecheck, full suite, stale-.js check, dep-audit, OpenAPI spectral lint — all green.

## Science-design questions for review

1. **`fac_strategic_value` in the assistant bundle.** Its enriched influence (0.167) is much lower than raw-ISL elasticity-derived influence (1.0). The PR correctly surfaces the enriched value. Worth confirming with Neil whether this lower number is the right user-facing signal — it reflects the genuine graph-based causal weight, but the magnitude drop will be visible.
2. **Confidence values universally drop.** PLoT's recomputed confidence (0.25 typical) is lower than raw-ISL `confidence` (0.36–0.38 typical) in both bundles. Same point as #1 — this is the intended provenance, but worth flagging the magnitude shift.

Both items are about transparency of an existing-but-hidden audit decision, not about whether the fix is correct.

## Follow-up (NOT in this PR)

A separate **"option assumptions to validate"** bucket in `m1_coaching` for intervention-target factors. Levers are still important assumptions (the user is committing to a value for each); they belong in a presentation surface, just not in the background uncertainty list. Suggested separate PR after this lands.

## Branch / process

- Branched cleanly from `origin/staging`.
- EVPI clamp branch (`claude-plot/evpi-non-negative-clamp`) is untouched.
- **Pausing for Paul / ChatGPT-led review. Do not merge or deploy.**

## Test plan

- [ ] Verify PLoT-recomputed `confidence` values look right in the staging UI for at least one assistant-shape and one hiring/expansion-shape bundle.
- [ ] Confirm intervention-target factors no longer appear in `m1_coaching.evidence_gaps`.
- [ ] Confirm `factor_sensitivity[]` array unchanged in the public response.
- [ ] Optional: post-merge, replay the same two bundles against staging and compare to the local-replay numbers above.

### DIFF
--- a/contracts/schemas/plot-response.schema.json
+++ b/contracts/schemas/plot-response.schema.json
@@ -334,7 +334,7 @@
         },
         "evidence_gaps": {
           "type": "array",
-          "description": "Top 3 factors where gathering evidence would most impact the decision",
+          "description": "Top background-uncertainty factors where gathering evidence would most reduce decision uncertainty (capped at 3). Excludes factors that are decision levers (i.e. any factor present in `options[i].interventions` keys); levers may be surfaced separately in a future `option_assumptions` bucket. The `confidence` and `influence` values published here are sourced from the same PLoT-recomputed signals used for the public `factor_sensitivity[]` array (plot_unified_v2 confidence; graph-merged `influence_score`) — never raw ISL `confidence` or `elasticity`.",
           "items": {
             "type": "object",
             "required": ["factor_id", "factor_label", "voi_score", "confidence", "influence"],
--- a/src/coaching/evidence-gaps.ts
+++ b/src/coaching/evidence-gaps.ts
@@ -12,6 +12,7 @@ import { computeEvpiPercentagePoints } from '../lib/evpi-emission.js';
 export function computeEvidenceGaps(inputs: CoachingInputs): EvidenceGap[] {
   const { factorSensitivity } = inputs;
   const thresholds = getThresholds();
+  const interventionTargetIds = inputs.interventionTargetIds ?? new Set<string>();
 
   if (factorSensitivity.length === 0) {
     return [];
@@ -33,10 +34,25 @@ export function computeEvidenceGaps(inputs: CoachingInputs): EvidenceGap[] {
     voi: f.normalisedImpact * (1 - (f.confidence ?? 0.5)),
   }));
 
-  // Top quartile with min floor from thresholds
-  const k = Math.min(Math.max(Math.ceil(factorSensitivity.length / 4), thresholds.evidence_gap_floor), factorSensitivity.length);
-  const gaps = withVoI
-    .filter((f) => f.importance_rank <= k && f.confidence < 0.7 && f.voi >= thresholds.evidence_gap_min_voi)
+  // Intervention-target factors are decision levers (set directly by options),
+  // not background uncertainties to validate, so they are excluded BEFORE the
+  // rank/quartile gate. Applying eligibility first prevents levers from
+  // consuming the top-k slots and hiding valid non-lever factors that ranked
+  // below them in the enriched importance order. Source of truth for lever
+  // identity is `options[i].interventions` keys (via
+  // `inputs.interventionTargetIds`), NOT raw ISL `zero_reason`. A future
+  // "option assumptions to validate" bucket may surface levers separately —
+  // documented follow-up; not in this PR.
+  const nonLevers = withVoI.filter((f) => !interventionTargetIds.has(f.node_id));
+  const k = Math.min(
+    Math.max(Math.ceil(nonLevers.length / 4), thresholds.evidence_gap_floor),
+    nonLevers.length,
+  );
+  const topByImportance = [...nonLevers]
+    .sort((a, b) => a.importance_rank - b.importance_rank)
+    .slice(0, k);
+  const gaps = topByImportance
+    .filter((f) => f.confidence < 0.7 && f.voi >= thresholds.evidence_gap_min_voi)
     .sort((a, b) => b.voi - a.voi)
     .slice(0, thresholds.evidence_gap_cap); // Cap from thresholds
 
--- a/src/coaching/key-drivers.ts
+++ b/src/coaching/key-drivers.ts
@@ -32,18 +32,25 @@ export function computeKeyDrivers(inputs: CoachingInputs): KeyDriver[] {
   // Use centralized normalised impact (same as evidence gaps)
   const normalisedImpactMap = computeNormalisedImpact(inputs.factorSensitivity);
 
-  // Sort by absolute influence score descending
-  const sorted = [...inputs.factorSensitivity].sort((a, b) => {
-    const aInfluence = Math.abs(a.elasticity ?? a.influence_score ?? 0);
-    const bInfluence = Math.abs(b.elasticity ?? b.influence_score ?? 0);
-    return bInfluence - aInfluence;
-  });
+  // Sort by absolute influence score descending.
+  // Provenance: prefer `influence_score` (canonical, graph-merged PLoT signal)
+  // over `elasticity` (raw ISL value) — same precedence as
+  // `computeNormalisedImpact`, so the published `rank`, `influence_score`, and
+  // `normalised_impact` fields are internally consistent. For graph-source
+  // enriched entries the two are equal; for ISL-only-append entries
+  // influence_score is the correct canonical signal.
+  const impactMagnitude = (f: typeof inputs.factorSensitivity[number]): number =>
+    Math.abs(f.influence_score ?? f.elasticity ?? 0);
+
+  const sorted = [...inputs.factorSensitivity].sort(
+    (a, b) => impactMagnitude(b) - impactMagnitude(a),
+  );
 
   // Take top 5
   const top5 = sorted.slice(0, 5);
 
   return top5.map((factor, index) => {
-    const influence = Math.abs(factor.elasticity ?? factor.influence_score ?? 0);
+    const influence = impactMagnitude(factor);
     const normalisedImpact = normalisedImpactMap.get(factor.node_id) ?? 0;
 
     return {
--- a/src/coaching/m1-coaching.ts
+++ b/src/coaching/m1-coaching.ts
@@ -6,7 +6,12 @@
  */
 
 import type { M1Coaching, Critique, Readiness } from './types.js';
-import type { EngineGraphV3, OptionV3, GoalConstraint } from '../types/engine-v3.js';
+import type {
+  EngineGraphV3,
+  FactorSensitivityResultV3,
+  GoalConstraint,
+  OptionV3,
+} from '../types/engine-v3.js';
 import { normaliseCoachingInputs } from './normalise-inputs.js';
 import { generateHeadlines, getFragileEdgeContext } from './headlines.js';
 import { computeEvidenceGaps } from './evidence-gaps.js';
@@ -27,6 +32,15 @@ import { generateExecutiveSummary } from './executive-summary.js';
  * @param logger Optional logger for warnings
  * @param repairsApplied Optional repairs from normaliser (_meta.repairs_applied)
  * @param ceeCritiques Optional critiques from CEE review
+ * @param goalConstraints Optional active goal constraints
+ * @param factorSensitivityEnriched Public/enriched `factor_sensitivity[]` array
+ *   (graph-merged influence, PLoT-recomputed confidence). When provided,
+ *   coaching consumes this as the canonical factor signal instead of raw ISL
+ *   `factor_sensitivity`. This keeps `m1_coaching.evidence_gaps` aligned with
+ *   the public payload's provenance — same confidence/influence values, same
+ *   intervention-override filtering. Falls back to raw ISL when omitted to
+ *   preserve backward-compatible behaviour for unit tests that mock ISL
+ *   responses directly.
  * @returns M1Coaching object or null if generation fails
  */
 export function generateM1Coaching(
@@ -46,7 +60,8 @@ export function generateM1Coaching(
     message: string;
     severity?: string;
   }>,
-  goalConstraints?: GoalConstraint[]
+  goalConstraints?: GoalConstraint[],
+  factorSensitivityEnriched?: FactorSensitivityResultV3[]
 ): M1Coaching | null {
   const startTime = performance.now();
 
@@ -55,7 +70,7 @@ export function generateM1Coaching(
     const thresholds = getThresholds();
 
     // Normalize inputs (required for all components)
-    const inputs = normaliseCoachingInputs(graph, options, islResult);
+    const inputs = normaliseCoachingInputs(graph, options, islResult, factorSensitivityEnriched);
 
     // Phase 2: Core Coaching (B1-B4)
     const storyHeadlines = safeCompute(() => generateHeadlines(inputs), {}, logger, 'headlines');
--- a/src/coaching/normalise-inputs.ts
+++ b/src/coaching/normalise-inputs.ts
@@ -152,6 +152,15 @@ export function normaliseCoachingInputs(
  * Compute normalised impact scores for factors (0-1 scale).
  * Used by both evidence gaps and key drivers to ensure consistency.
  *
+ * Provenance: prefer `influence_score` (canonical, graph-merged signal carried
+ * on enriched FactorSensitivityResultV3 entries) over `elasticity` (raw ISL
+ * value preserved on enriched ISL-only-append entries and on legacy raw-ISL
+ * inputs). For graph-source enriched entries `influence_score === elasticity`
+ * so the choice is moot; for ISL-only-append entries the two can differ and
+ * influence_score is the right canonical signal to propagate into the public
+ * `m1_coaching.evidence_gaps[].influence` field — matching the same provenance
+ * promise the public `factor_sensitivity[].influence_score` carries.
+ *
  * @param factors Factor sensitivity data
  * @returns Map of factor ID to normalised impact (0-1)
  */
@@ -162,15 +171,10 @@ export function computeNormalisedImpact(
     return new Map();
   }
 
-  const maxImpact = Math.max(
-    ...factors.map((f) => Math.abs(f.elasticity ?? f.influence_score ?? 0)),
-    0.001
-  );
+  const impactOf = (f: NormalisedFactorSensitivity): number =>
+    Math.abs(f.influence_score ?? f.elasticity ?? 0);
 
-  return new Map(
-    factors.map((f) => [
-      f.node_id,
-      Math.abs(f.elasticity ?? f.influence_score ?? 0) / maxImpact,
-    ])
-  );
+  const maxImpact = Math.max(...factors.map(impactOf), 0.001);
+
+  return new Map(factors.map((f) => [f.node_id, impactOf(f) / maxImpact]));
 }
--- a/src/coaching/normalise-inputs.ts
+++ b/src/coaching/normalise-inputs.ts
@@ -14,35 +14,60 @@ import type {
   NormalisedOption,
   NormalisedRobustness,
 } from './types.js';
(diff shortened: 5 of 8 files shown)


Talchain:
Thanks for the review. Pushed commit `bd222b8` addressing P1 + P2 (impact provenance + schema description). Untracked diagnostic script kept out of the PR per your note.

### P1 — influence/VOI provenance bug (real, latent in both staging bundles)

You were right. `computeNormalisedImpact` preferred `elasticity ?? influence_score`, so when an enriched entry carried both fields (true for ISL-only-append entries in `mergeIslConfidenceIntoGraphFactors`, where ISL's raw elasticity is preserved via the `...islFCleaned` spread), the published `evidence_gap.influence` was being driven by raw-ISL elasticity rather than the graph-merged `influence_score`.

The bug was **latent in both Task 3 bundles** — confirmed: all enriched entries in `c1abcea5` and `c952d8d6` are `source: 'graph'`, where `elasticity === influence_score` by construction. The bug would have manifested the moment the merge appended any ISL-only factor (different graph vs ISL ranking).

Fix in `src/coaching/normalise-inputs.ts`: `computeNormalisedImpact` now prefers `influence_score ?? elasticity`. Scope decision below.

### P2 (a) — scope of the threading change

You're right that threading enriched factor_sensitivity into `CoachingInputs` affects more than `evidence_gaps`. Going through each consumer:

- `key-drivers.ts` (line 36) reads `elasticity ?? influence_score` directly — same legacy ordering. I haven't changed it directly, but it also calls `computeNormalisedImpact` indirectly via `normalised_impact`, so the helper change there is consistent.
- `critiques.ts` (line 45) reads `f.elasticity ?? 0` with no influence_score fallback. For graph-source entries `elasticity === influence_score`, so DOMINANT_FACTOR/INFLUENTIAL_EXTERNALS still fire correctly. For ISL-only-append entries this is a residual gap — happy to take that as a follow-up since it's outside this PR's evidence-gap scope.
- `headlines.ts` (line 92) has its own inline `elasticity ?? influence_score` for VOI. Same residual gap; same follow-up reasoning.
- `executive-summary.ts` — no direct elasticity read.

I've kept this PR narrow: `computeNormalisedImpact` helper + the explicit lever filter. The broader alignment of `critiques.ts` / `headlines.ts` should be a follow-up PR with its own scope justification. Flagging this in the commit message so the audit trail is honest.

### P2 (b) — schema description

`contracts/schemas/plot-response.schema.json` `evidence_gaps[]` description expanded to make three contracts explicit: (a) background uncertainties, not levers; (b) levers excluded via `options[].interventions` filter, may surface separately in a future `option_assumptions` bucket; (c) confidence/influence sourced from the same PLoT-recomputed signals as the public `factor_sensitivity[]` (never raw ISL).

### P2 (c) — missing impact-provenance test

Two new tests added in `tests/coaching/evidence-gaps-provenance.test.ts`:

1. `end-to-end gap.influence is computed from enriched influence_score, not raw-ISL elasticity` — fixture with `{elasticity: 0.05, influence_score: 0.9}`, asserts `gap.influence === 1.0`. Locks the provenance promise at the public emission surface (your specific request).
2. `with two enriched factors whose elasticity-vs-influence_score orderings differ, ranking follows influence_score` — two factors where elasticity-based ordering would invert influence_score-based ordering. Top-1 must follow `influence_score`.

A route-level test asserting parity between `factor_sensitivity[]` and `m1_coaching.evidence_gaps[]` for the same factor is the right next step, but it needs a more involved fixture (full /v2/run wiring). Flagging as a follow-up alongside the `critiques.ts` / `headlines.ts` alignment.

### Untracked diagnostic script

`scripts/diagnostics/before-after-evidence-gaps.ts` stays out of this PR. You're right about the contradiction — it derived "intervention targets" from raw ISL `zero_reason` (which the PR explicitly rejects as the canonical source). It was a one-off verification tool, not production code. I'll either delete it locally or rewrite it to derive from `options[].interventions` if it becomes a permanent diagnostic.

### Verification

- `npx tsc --noEmit` — clean.
- `npx vitest run tests/coaching/evidence-gaps-provenance.test.ts tests/coaching/m1-coaching.test.ts tests/3c-factor-sensitivity-enrichment.test.ts tests/b8-8-3c-field-segregation.test.ts` — **104/104 pass**.
- `bash scripts/pre-push-validate.sh` — **4685 tests pass / 25 skipped / 0 fail**, tsc clean, OpenAPI spectral lint clean.

Re-running the staging-replay against `c1abcea5` and `c952d8d6` with the corrected helper produces identical evidence_gaps to the previous run (as expected — both bundles are graph-source so the latent bug doesn't bite there). The fix is a guard against future bundles that go through the ISL-only-append path.

Talchain:
Round 2 review acknowledged. Pushed `e6f4bd1` addressing the key-drivers consistency bug. Per-point response below.

### P1 (key_drivers internal inconsistency) — fixed in `e6f4bd1`

You were exactly right. After the previous commit flipped `computeNormalisedImpact` to prefer `influence_score`, `key-drivers.ts` was left sorting and publishing the `influence_score` field via `Math.abs(elasticity ?? influence_score)` — so a single `key_drivers[]` entry could ship an elasticity-ordered `rank` and `influence_score` alongside an influence-score-ordered `normalised_impact`. My earlier "consistent" claim was wrong; the helper aligned one field but not the other two on the same entry.

Fix: hoisted the impact-magnitude into a local helper inside `computeKeyDrivers`:

```ts
const impactMagnitude = (f) => Math.abs(f.influence_score ?? f.elasticity ?? 0);
```

Used for both the sort key and the published `influence_score`. Now `rank`, `influence_score`, and `normalised_impact` within any single entry all derive from the same canonical signal.

Two tests added in `tests/coaching/evidence-gaps-provenance.test.ts`:

- `rank ordering follows influence_score precedence, not elasticity` — two factors with rank-inverting elasticity vs influence_score. Pre-fix the elasticity-leader ranked top; post-fix the influence_score-leader ranks top.
- `within a single key_drivers entry, influence_score and normalised_impact agree on the same source` — end-to-end consistency assertion: the entry whose `normalised_impact === 1.0` must have the max published `influence_score`. Contradiction-free invariant.

### P2 (helper changes raw-ISL fallback behavior) — scope decision stands, documented

Acknowledged: the helper change is global, not enriched-only. I considered your alternative ("clear legacy `elasticity` on the enriched path" or "canonical impact field") and rejected both because:

- **Clearing elasticity on the enriched path** would force a parallel change in `critiques.ts` (line 45: `Math.abs(f.elasticity ?? 0)` with no influence_score fallback) so DOMINANT_FACTOR/INFLUENTIAL_EXTERNALS continue to fire on enriched entries. That widens scope to a third coaching component, plus its tests.
- **A new "canonical_impact" field** is a schema/type surface expansion that introduces a new contract for one bug fix.

The current approach — change `computeNormalisedImpact` precedence and align `key-drivers.ts` — touches two places, both already aligned on the same provenance promise. Behaviour difference in the legacy raw-ISL fallback path:

- Production never uses legacy fallback (run.ts always supplies enriched at line 4309).
- Unit tests typically use `elasticity === influence_score` so they don't observe the change.
- For unit tests that DO use unequal values, the change makes the legacy path "prefer the more canonical signal", which is a strict improvement in test realism even if it's not strict backward compatibility.

If you (or Neil) prefer the strict-fallback-preservation invariant, I can add a `from_enriched: true` marker on `NormalisedFactorSensitivity` and branch the helper on it. Quick to do; ask if you want it.

### P2 (OpenAPI doc gap) — confirmed not present; nothing to update

Direct grep confirmed:

```
$ grep -n "evidence_gaps\|m1_coaching" contracts/openapi.yaml
(no matches)
```

`m1_coaching` is documented exclusively in `contracts/schemas/plot-response.schema.json`. The OpenAPI spec doesn't describe the coaching block at all on this branch. So the single schema description update is sufficient for v1. If `m1_coaching` ever gets documented in OpenAPI separately, the same description text should be copied. Flagging in the file change so future-me notices.

### P2 (single-factor test is mostly redundant) — accepted; framing fixed

You're right. The test self-acknowledges in its comment ("would have been 1 if computed from elasticity too"), but I shouldn't have framed it as "proving" provenance — both that test and the two-factor inversion test are now in the file, and only the two-factor test is the actual regression guard. The single-factor test stays as end-to-end shape coverage (display string, voi_score scalar). Updated commit message phrasing reflects this.

### P2 (display-honesty plan not in repo / not auditable) — flagging for decision

Correct. The plan file lives at `~/.claude/plans/display-honesty-plan-A-B-C.md` — my personal note-taking location, not in the repo. The "updated to v2" claim is unauditable to external reviewers from the PR alone.

This is intentional for now (the plan is for the user to review before any code is written), but I see your point. Three options:

1. **Move into the repo** — e.g., `docs/plans/display-honesty.md`. Auditable but creates a doc surface that needs maintenance.
2. **Reference in PR body** — paste the plan content (or a link, if linkable) into PR #166's body or a follow-up draft PR.
3. **Leave as-is** — the plan is for Paul; reviewers don't need to verify it line-by-line at PR-#166 time because the plan describes future PRs B/A/C, not this PR.

I lean toward (3) for now and (1) when work on B/A/C actually starts (the plan becomes a working doc at that point). Happy to flip if you want it visible now — just say which option.

### Verification

- `npx tsc --noEmit` — clean.
- `npx vitest run tests/coaching/evidence-gaps-provenance.test.ts tests/coaching/m1-coaching.test.ts tests/3c-factor-sensitivity-enrichment.test.ts` — **101/101 pass**.
- `bash scripts/pre-push-validate.sh` — **4687 tests pass / 25 skipped / 0 fail**, tsc clean, OpenAPI spectral lint clean.

The same-bundle replay against `c1abcea5` and `c952d8d6` still produces identical outputs (both bundles are graph-source where `elasticity === influence_score`, so this layer's behaviour is unchanged on the bundles). The fix is a guard against future ISL-only-append cases for both `evidence_gaps[]` and `key_drivers[]`.

Talchain:
## Follow-ups (out of scope for this PR)

`critiques.ts` and `headlines.ts` still read elasticity directly; review in a separate coaching-provenance sweep if those surfaces are promoted or shown alongside evidence-gap/factor-sensitivity values.

Talchain:
Round 3 review acknowledged. Merge recommendation noted. Pushed `90ebf96` with two small, in-scope follow-ups; the rest are logged for separate work.

### Acted on in this PR (commit `90ebf96`)

**P2 — fallback-comment honesty.** Updated the docstring on `enrichedFactorSensitivity` in `src/coaching/normalise-inputs.ts` to say the legacy path preserves the **shape** of raw-ISL coaching inputs but **numeric semantics may differ** for inputs that supply both `elasticity` and `influence_score` with different values (because `computeNormalisedImpact` now globally prefers `influence_score`). Matches the public `factor_sensitivity[]` contract; intentional.

**P3 — `Array.isArray` guard on lever derivation.** Added `&& !Array.isArray(interventions)` to the lever-set loop. Without it, an array-shaped `interventions` (which is a `typeof === 'object'`) would have `Object.keys(arr)` produce positional indices (`"0"`, `"1"`, …) as lever ids. Route schema enforces plain-object in production, so this only matters for malformed direct callers, but it's cheap defence-in-depth. New test covers array / null / string / number shapes.

### Not acted on (deferred per user instruction + reviewer concurrence)

**P2 — `headlines.ts` raw-elasticity precedence (lines 89, 150).** User-facing coaching copy, but only diverges from evidence_gaps for ISL-only-append factors (graph-source has `elasticity === influence_score`). Reviewer agrees this should not block the PoC. Already logged in the follow-ups note ([comment #4460997327](https://github.com/Talchain/plot-lite-service/pull/166#issuecomment-4460997327)).

**P2 — `critiques.ts` elasticity-based heuristics (lines 45, 70, 78).** Same rationale. Same follow-up note.

**P3 — Duplicate factor IDs.** Reviewer themselves flag as pre-existing weakness not introduced by this PR. Added to the follow-ups note (see PR comment update below) — wants a separate dedupe-guard change with its own tests, not a quiet add here.

### Updated follow-ups note

> `critiques.ts` and `headlines.ts` still read elasticity directly; review in a separate coaching-provenance sweep if those surfaces are promoted or shown alongside evidence-gap/factor-sensitivity values.
>
> Duplicate factor IDs reaching coaching: `computeNormalisedImpact`'s Map collapses but evidence-gap iteration still emits multiple rows. Pre-existing weakness; not introduced by this PR. Worth a small dedupe guard on the normalised factor list when the upstream invariants are formalised.
>
> Route-level parity test: `/v2/run` integration assertion that `factor_sensitivity[]` and `m1_coaching.evidence_gaps[]` agree on `confidence`/`influence` for the same factor.
>
> Future "option assumptions to validate" bucket for intervention-target factors: needs its own design brief, not bolted onto this PR.

### Verification

- `npx tsc --noEmit` — clean.
- `npx vitest run` on the focused coaching files — **80/80 pass** (1 new test).
- `bash scripts/pre-push-validate.sh` — **4688 tests pass / 25 skipped / 0 fail**, tsc clean, OpenAPI spectral lint clean.

Ready for merge per your recommendation. Not merging or deploying without explicit go-ahead.

PR #185, opened 2026-06-10
fix(flip-thresholds): forward resolved analysis seed to flip-threshold probes (intentional CRN)

Talchain:
## Summary
Flip-threshold probe requests were built without a `seed`, so ISL fell back to a graph-hash-derived seed that ignores the request seed. Because that fallback hash excludes the probed value, all probe points shared draws by accident — common random numbers, but in a **different Monte Carlo world** than the displayed base analysis (probe baseline margin diverged from the main-call margin). This forwards the **same resolved seed** PLoT already sends to the main ISL robustness call onto every flip-threshold probe.

## Change (single area)
`src/analysis/flip-thresholds.ts` · `createISLInferenceFn`:
- add `seed?: string | number` to the `originalRequest` type;
- include `seed` in the per-probe `modifiedRequest`, **omitting the key entirely when no seed is present** so seedless callers keep ISL's graph-hash default.

The `/v2/run` call site already passes `islRequest`, whose seed is `plotSeedUsed = resolveSeed(providedSeed, graph)` (the explicit request seed, or the PLoT-derived seed when omitted) — so **no route change**. The coaching flip-threshold path only selects candidates and builds no ISL requests, so it is untouched. **There is no second flip-probe ISL request path.**

## Behaviour
- Probe world aligns with base analysis: same seed → probe baseline margin matches the main-call margin within representation/rounding limits.
- Intentional CRN preserved: **one constant seed across all probe points** within a single search (never per-probe, never the graph-hash fallback).

## Acceptance signal (durable)
The durable acceptance signal is **probe baseline margin matching the main-call margin under the resolved seed** — not any specific flip value. Verified locally against staging ISL with the `delivery_gap` artefact (seed 4242): probe baseline margin moved **0.309 → 0.346** to match the main-call margin; `max_probe_delta` moved −1.052 → −1.129 (the probe MC world genuinely changed).

**`delivery_gap.flip_value = 6.055` is not a fixed contract.** It happened to remain `6.055` at `n=1000` only because the crossing landed at the same bisection midpoint within ~0.1-story-point resolution — but it is now produced under the **aligned** seed. **Future acceptance should re-capture the current artefact value rather than hard-code the old number.**

## Out of scope (unchanged)
Response hashing, timeout behaviour, default `n_samples`, ISL, CEE, DGAI/UI, prompts/PMS, schemas/contracts. No Branch A replay.

## Tests
- `tests/analysis/flip-thresholds.test.ts`: probe forwards the resolved seed; forwards a PLoT-derived (omitted-then-resolved) seed; one constant seed across all probe points (CRN); omits seed when absent (graph-hash default preserved); no mutation of the original request; same seed → identical flip result; different seeds → different result / different probe seeds.
- `tests/v2-run.flip-probe-seed.contract.test.ts` (new): full `/v2/run` path — explicit and omitted-then-derived seeds both reach every `__flip` probe and equal `meta.seed_used`; different seeds → different probe seeds.

## Reviewer checklist (Codex focus)
- [ ] No hidden second flip-probe ISL request path (only `createISLInferenceFn` builds probe requests; coaching path selects candidates only).
- [ ] Probes receive the **resolved** seed (`plotSeedUsed`), not merely the raw optional request seed.
- [ ] Omitted-seed path forwards the **derived** resolved seed (covered by the contract test's omitted-then-derived case).
- [ ] Seedless callers with truly no seed preserve ISL's graph-hash default (key omitted).
- [ ] Probe points share one seed within a single search (intentional CRN).
- [ ] Response hash, timeout behaviour, default `n_samples`, ISL, CEE, DGAI/UI unchanged.

### DIFF
--- a/src/analysis/flip-thresholds.ts
+++ b/src/analysis/flip-thresholds.ts
@@ -538,11 +538,27 @@ function roundTo4(value: number): number {
  *    to the probe value — the field ISL's comparison reads as the sampling mean
  * 2. Keeps the target factor's parameter_uncertainties.mean aligned to the same
  *    value (contract honesty; the comparison shadows this field)
- * 3. Calls ISL via the service's callAnalysisEndpoint
- * 4. Returns the option comparison results
+ * 3. Forwards the resolved analysis seed (the same seed sent to the main ISL
+ *    robustness call) on every probe request
+ * 4. Calls ISL via the service's callAnalysisEndpoint
+ * 5. Returns the option comparison results
+ *
+ * Seed forwarding + intentional common random numbers: every probe in a single
+ * flip search receives `originalRequest.seed` — the resolved seed PLoT sends to
+ * the main analysis call (the explicit request seed, or the seed PLoT derived
+ * from the canonical request/graph when omitted). Holding the seed constant
+ * across probe points means ISL's PCG64 sampler draws the *same* edge
+ * configurations and factor-noise z-values at every probe; only the probed
+ * factor's sampling mean shifts. That preserves the deterministic
+ * common-random-numbers smoothness the bisection relies on, while aligning the
+ * probe world with the displayed base analysis (same seed → probe baseline
+ * margin matches the main-call margin within representation/rounding limits).
+ * We deliberately do NOT vary the seed per probe and do NOT fall back to ISL's
+ * graph-hash seed (which ignores the request seed).
  *
  * @param callAnalysis - Function that calls ISL analysis endpoint
- * @param originalRequest - The original ISL robustness request
+ * @param originalRequest - The original ISL robustness request (carries the
+ *   resolved seed forwarded to the main analysis call)
  * @param requestId - Request ID for tracing (suffixed with flip search context)
  * @returns ISLInferenceFn callback
  */
@@ -554,6 +570,10 @@ export function createISLInferenceFn(
     goal_node_id: string;
     n_samples?: number;
     parameter_uncertainties?: Array<{ node_id: string; distribution: string; mean: number; std: number }>;
+    // Resolved seed sent to the main ISL analysis call (explicit request seed,
+    // or PLoT-derived seed when omitted). Forwarded verbatim on every probe so
+    // common random numbers stay intentional and aligned with the base analysis.
+    seed?: string | number;
   },
   requestId: string
 ): ISLInferenceFn {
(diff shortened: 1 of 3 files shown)


Talchain:
## Review response (4th pass)

Thanks — triaged each point. No PR-specific blocker found again; keeping the diff at exactly the 3 approved files (no production-code change since `8a27694`). Verdicts:

### Live ISL actually consumes `seed` (closes the "mocked ISL" test gap)
The automated suite mocks ISL deliberately to stay hermetic, but live `/api/v1/robustness/analyze/v2` consumption was verified read-only against **staging ISL `4f5bf77`**:

- **ISL-direct** (comparison-only, n=1000, same payload): seed `4242` twice → identical wins `[0.231, 0.577, 0.192]`; seed `777` → `[0.249, 0.571, 0.18]`; seedless → `[0.246, 0.555, 0.199]`. → ISL is deterministic per seed and the seed is **consumed, not inert**.
- **End-to-end A/B** (local PLoT-with-fix → staging ISL, `delivery_gap` artefact, seed 4242): probe baseline margin moved **0.309 → 0.346** to match the main-call margin; `max_probe_delta` −1.052 → −1.129. → the forwarded seed changes probe behaviour against the *real* sampler exactly as intended.

I'm not adding a live-ISL test to the committed suite on purpose: it would make CI depend on staging availability + API keys and reintroduce the network flakiness we already see in spawn-based tests. The manual evidence above is the verification.

### Not changing in this PR (out of scope / non-goals), tracked as follow-up
Consolidated into one ISL-contract-hygiene follow-up:
- shared `contracts/schemas/isl-request.schema.json` doesn't declare `seed` (permissive via `additionalProperties: true`, so no runtime rejection — pure drift vs the live `ISLRobustnessRequestV3`);
- stale **unused** `ISLRobustnessAnalyzeV2Request` interface omits `seed`;
- `createISLInferenceFn`'s inline `originalRequest` shape could be `Pick<ISLRobustnessRequestV3, …>` to prevent future field drift.

These are contract/schema/type files (explicit non-goals for this narrow PR) and the type refactor risks rippling into the probe's `any`-based graph mutation + tests, so they're deferred rather than bundled here.

### Confirmations (no action)
- **`response_hash` unchanged** — the resolved seed is already part of request canonicalisation, and flip-threshold diagnostics are excluded from the response hash. Independent of this diff. ✅
- **Empty-string seed** — the new unit test documents *layer* behaviour (verbatim forwarding) only; whether `''` is a valid product seed is an existing `/v2/run` route/schema question, not decided here. ✅
- **No second live `__flip` builder** — `createISLInferenceFn` is the only one; the CEE orchestrator fallback isn't exercised because `/v2/run` passes pre-resolved flip data. ✅
- **Downstream flip values may shift** — intended; `delivery_gap.flip_value = 6.055` is explicitly **not** a contract (it only coincidentally held at n=1000). Consumers should re-capture, not preserve old goldens.

### Merge-gating CI policy (for Paul)
`audit` (fast-uri/fastify advisories) and `gates (windows-latest)` (checkout fails on the pre-existing invalid path `tools/sdk-smoke:python.mjs`) fail on **every** PR in this repo, unrelated to this change. `test`, `verify`, `guard`, `smoke`, `safety`, and the Linux/macOS gates pass. If branch protection requires all checks green, these two need a fix or waiver as a repo-policy decision — out of scope for this PR.

Deferred investigation items (timeout honesty, probe-`n_samples` decoupling, hash-v7 with `n_samples`, seed-sweep tooling, semi-analytic threshold) remain tracked separately and intentionally out of this PR.

PR #212, opened 2026-07-10
feat(v2): typed failure envelope for /v2/run — discriminating failure codes, revived ISL-422 passthrough, error.v1 unknown-field 400 (fragility gaps 1/2/3/6/8)

Talchain:
# Typed failure envelope for /v2/run (fragility gaps 1/2/3/6/8)

PLoT half of `parallel-briefs/FRAGILITY-PAIR-FINDINGS-2026-07-10.md` (§seam split). RED-first:
`tests/v2-typed-failure-envelope.test.ts` committed RED (7 fail / 1 pass on base `1eaedd9f`),
now 8/8 GREEN.

## What changed, per gap

| Gap | Before | After |
|---|---|---|
| **2** | Every ISL infrastructure failure collapsed to `ISL_CALL_FAILED`; the discriminating codes computed in `integrations/isl/index.ts` were discarded at the route; `status_reason` always `'ISL analysis failed'`; `islStatusCode` hardcoded 500 | Discriminating critique (`ISL_TIMEOUT` / `ISL_NETWORK_ERROR` / `ISL_ERROR` / `ISL_REJECTED`) carried **alongside** the legacy `ISL_CALL_FAILED` critique (back-compat kept); honest per-class `status_reason`; `islStatusCode` = ISL's real HTTP status |
| **3** | ISL's structured 422 critiques were dead code — `callAnalysisEndpoint` swallows `ISLHttpError`, so the route's `is422()` branch never fired; ISL validation detail flattened into gap 2 | `callAnalysisEndpoint` keeps its no-throw contract but attaches `status` + normalised `critiques` to the returned error; route maps them through `mapISLCritiquesToV2` → same 422-blocked contract as the throw path (which stays as belt). A 422 **without** parseable critiques returns failed(200) with the `ISL_REJECTED` discriminator — never blocked-with-empty-critiques |
| **1** | Internal PLoT exception → HTTP 200, `analysis_status:'failed'`, **empty `critiques[]`** | Typed `PLOT_INTERNAL_ERROR` critique (source `engine`), generic user-safe copy; raw exception goes to the log, never the wire; `meta.request_id` carried |
[...]

PR #213, opened 2026-07-10
feat(v2): evidence provenance — response_content_hash + CEE digests/chain legs (ROADMAP 2.13)

Talchain:
# Evidence-provenance completion for /v2/run (ROADMAP 2.13)

Closes the PLoT half of the 2.13 residual gaps identified by the 2026-07-10 six-explorer sweep
(the row was otherwise largely landed by Lane PLoT-R3: `_meta.evidence` digests, downstream
capture, request-id chain, UI `evidence_capture`). RED-first: `tests/v2-evidence-provenance.test.ts`
committed RED (module-absent), now 9/9 GREEN.

## The four gaps closed

| Gap | Before | After |
|---|---|---|
| **A — content hash** | No deterministic hash of what PLoT RETURNED existed. `response_hash` hashes the canonicalised REQUEST (by design — determinism token; the UI freshness gate keys on `response_hash === results.hash`). The only body hash was the `x-olumi-response-hash` header: header-only, covers latency/timestamps → non-reproducible | Additive **`_meta.response_content_hash`** = `rch_v1:<sha256-16>` over the public semantic surface minus the exported volatile set (`src/util/response-content-hash.ts`), with the determinism-replay natural-key array sorts. Attached after the body is built (never hashes itself); success responses only (`V2RunError` has no `_meta`). Replay-stable (determinism-replay suite green). `response_hash` untouched; its request-canonical semantics now DOCUMENTED as by-design in the OpenAPI endpoint description (the row's "fix or document" → document + add distinctly-named content hash) |
| **B — zero-mismatch assertions** | "Zero hash mismatches" was never asserted: nothing compared header-vs-body or recorded-digest-vs-actual-bytes | Within-run assertions in the new suite: `x-olumi-response-hash` header == independent body recompute; `response_content_hash` independently recomputable from the body (exclusion list is exported precisely so external verifiers can recompute) |
| **C — CEE byte digests** | `cee/client.ts` never computed `PayloadDigest`s — diligence-grade byte digests were ISL-only | All 3 CEE call sites now read raw text-then-parse (matching the ISL client) and record `requestDigest`/`responseDigest` over the exact wire bytes + the echoed `x-request-id` |
| **D — CEE chain leg** | `request_id_chain` was PLoT→ISL only | Additive `cee`/`cee_echoed` legs, derived from the downstream tracker at response build. **`all_match`/`chain_complete` stay computed over the original ISL four** (Brief 4 header spec back-compat) — pinned by test: a divergent CEE echo cannot flip them |

## Deliberate pin updates (each verified, none papering)

- `tests/b3-numeric-invariance.test.ts`: `response_content_hash` added to the ambient mask — it is
  a derived function of the two B3 fields that test deliberately lets differ; its own correctness
  is pinned in the new suite.
- ISL-V2 golden regenerated TWICE, both diffs fully explained: (1) the three designed additive
  fields (cee legs ×2 locations + content hash); (2) after the golden itself caught `meta.build`
  (deployed git SHA) inside the hash — every commit flipped the hash on identical content →
  `build` + `feature_flags` moved to the exclusion list as deployment/environment provenance,
  golden regenerated, delta = hash value only. This is the golden doing exactly its job.

## Consciously out of scope

- **State-selection trace** — pure UI-half (extend `analysis_state_source`, fill the stubbed
  `cross_surface_events: []`); routed to A2 via the orchestrator (sweep report has the seam).
- **`UI_CANONICAL_META` staging flip** — `_meta.payloads.isl_request/response` (full sanitized
  bodies) stay gated; un-gating is an env/exposure decision for the orchestrator, not this lane.
  The always-on digest surface (`_meta.evidence`) is unaffected.
- CEE-side echo: CEE may not echo `x-request-id` today — `cee_echoed` reports the honest null
  until it does (no CEE change required for the field to be truthful).

## Verification

- RED (commit `e0408ca`, module-absent) → GREEN 9/9; affected set green: determinism-replay,
  b3-numeric-invariance, evidence-capture, auto-constraint-fallback, cil-phase02 (the latter two
  needed `npm run build` first — known fresh-worktree dist/ env issue).
- Full `scripts/pre-push-validate.sh` PASSED (5,45x tests). Two intermediate gate runs flaked on
  DIFFERENT server-spawning files (v2-determinism, counterfactual.zero-baseline) — each passes in
  isolation, neither touched by this diff; known load-flake class (ROADMAP hygiene row).
- CI: expect the standing reds only — `audit`, `gates (windows-latest)`, and `validate-structure`
  (spec-touching PRs only; 37 pre-existing Redocly errors, proven identical on base — see PR #212
  comment for the evidence pattern).

### DIFF
--- a/contracts/openapi.yaml
+++ b/contracts/openapi.yaml
@@ -4031,7 +4031,15 @@ paths:
         - `drivers_status`: 'computed' | 'unavailable' | 'skipped' | 'error'
 
         **Response Hash**:
-        - `response_hash`: 16-char hex hash computed from semantic fields only (for determinism)
+        - `response_hash`: 16-char hex hash computed from semantic fields only (for determinism).
+          BY DESIGN this hashes the canonicalised REQUEST (graph, options, seed, identifiability
+          inputs) — a determinism/reproducibility token, NOT a hash of the response content.
+          Consumers use `response_hash` equality to prove "same analysis inputs"; it does not
+          verify response integrity.
+        - `_meta.response_content_hash` (additive, 2.13): "rch_v1:"-prefixed 16-char hex hash of
+          the response CONTENT — the public semantic surface minus volatile fields (timings,
+          timestamps, per-request ids, the `_meta` subtree). Deterministic across replays of the
+          same request and independently recomputable from the body. Success responses only.
 
         **Seed Handling**:
         - Accept string (canonical) or number (legacy)
--- a/src/cee/client.ts
+++ b/src/cee/client.ts
@@ -17,7 +17,7 @@ import type {
 import { runDecisionReviewViaSdk, type EvidenceHelperItem } from './orchestrator.js';
 import { isFlagOn } from './codes.js';
 import { computeOlumiHash } from '../util/canonical.js';
-import { recordDownstreamCall, sanitizePayloadForDebug } from '../util/downstream-tracker.js';
+import { recordDownstreamCall, sanitizePayloadForDebug, computePayloadDigest } from '../util/downstream-tracker.js';
 import {
   CEE_TIMEOUT_MS as CENTRAL_CEE_TIMEOUT_MS,
   CEE_DECISION_REVIEW_TIMEOUT_MS as CENTRAL_DECISION_REVIEW_TIMEOUT_MS,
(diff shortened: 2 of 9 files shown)


Talchain:
CI failure attribution: the three reds are the standing set — `audit`, `gates (windows-latest)` (every PR), and `validate-structure` (spec-touching PRs only). Re-verified for THIS branch: Redocly problem-set diff vs base staging is EMPTY (identical 37 errors / 39 warnings, same locations — same evidence pattern as the #212 comment). This PR's spec change is doc-prose only.

PR #250, opened 2026-07-22
fix(optimise): honesty fixes — kill synthetic bands, 400 on invalid graph, method disclosure (A1 locus, R-9)

Talchain:
## What (3 commits, one per fix — /v1/optimise, the ruled A1 "best combination" locus; ZERO callers exist)
- **e0cbb28 — R2: synthetic bands KILLED.** `utility.p10/p50/p90` were `expected*0.9 / ==expected / *1.1` — fabricated bands presented as computed. Deeper finding: `expected` is a greedy-additive sum of per-action median deltas from DISTINCT kernel runs over DIFFERENT graphs — no single MC distribution backs it, so NO honest band exists without changing the estimator (rowed: exact/joint upgrade). Response now `utility:{expected}` only. Complete reader manifest (adversarially verified, positive-controlled): zero readers of the killed bands anywhere (repo, both SDKs, fixtures, examples, cross-repo indicative sweep).
- **8bb236d — R3: missing/invalid `graph` → clean 400** (`BAD_INPUT` via the repo's `replyWithAppError` envelope), guarded before any graph access; positive control pins no over-rejection (well-formed empty graph still 200). Previously an unguarded deref → 500.
- **df058f5 — honest method disclosure (additive):** `method:"greedy_independent_v1"` + `action_semantics:"edge_weight_scaling"` — the wire must not imply joint/exact optimisation or Pearl do() semantics the route doesn't have. Both markers verified truthful at the bytes by the adversarial pass.

## Evidence (full gate)
RED-first 7 failed / 1 passed at base (digit-reproduced independently) · per-fix mutation isolation in throwaway worktrees (2/4+control/1 RED, cross-isolation clean, independently re-run) · suite 553 files / 5890 tests / 0 failed · typecheck exit 0 · ZERO golden deltas · openapi untouched and spec-legal (utility declared opaque; marker documentation = follow-up) · **Adversarial verdict: SAFE** (`acceptance-evidence/science-step1-2026-07-22/a1-optimise-fixes/ADVERSARIAL.md`).

## Rowed (pre-existing at base, surfaced by the adversarial — not in this diff)
[...]

PR #251, opened 2026-07-22
chore(optimise): /simplify cleanup — guard simplification, label co-location, test fold (zero-wire)

Talchain:
## What (1 commit, behavior-preserving — /simplify apply set, 22 Jul)
- Drop the redundant `typeof body.graph !== 'object'` guard clause (no value class it uniquely catches — falsy fails `!body.graph`, truthy primitives fail the `Array.isArray` checks with the same 400). All 5 validation tests pass unchanged.
- Drop two no-op `as const` on the disclosure consts (literal types already inferred).
- Lift the inline `meta.solver` string `'greedy_kernel_v1'` into `OPTIMISE_META_SOLVER`, co-located with `OPTIMISE_METHOD` + a cross-reference comment: both label the SAME greedy algorithm and must move together (mirror-drift made visible; **both wire values unchanged** — actual unification is a rowed item).
- Fold the wholly-subsumed second honesty-bands test into the first (keys-equality already proves band absence; single payload-independent return site). Net −1 test.

## Gate
Suite 5890→5889 (delta = exactly the fold), 0 failed · typecheck exit 0 · **ZERO golden deltas** (fixtures replay "All fixtures match") · full adjudication record: `acceptance-evidence/science-step1-2026-07-22/SIMPLIFY-FINDINGS-2026-07-22.md` (4-angle Fable review, apply/row/skip sets).

Expected pre-existing CI reds (verified NOT-this-diff): gates(windows-latest) colon-path · `audit` fast-uri advisory GHSA-v2hh-gcrm-f6hx (reproduced on clean base; dedicated fix outstanding).

### DIFF
--- a/src/routes/v1/optimise.ts
+++ b/src/routes/v1/optimise.ts
@@ -63,8 +63,13 @@ async function evaluateUtility(graph: any, objective: any, seed: number): Promis
 // - edge_weight_scaling: an action's do:[{node_id,set_to}] MULTIPLIES the weight of every
 //   outgoing edge of node_id by set_to. It does NOT set the node's value (no Pearl
 //   do-operator); a leaf/no-outgoing node is a silent no-op. Flagged for Neil doctrine.
-const OPTIMISE_METHOD = 'greedy_independent_v1' as const;
-const OPTIMISE_ACTION_SEMANTICS = 'edge_weight_scaling' as const;
+const OPTIMISE_METHOD = 'greedy_independent_v1';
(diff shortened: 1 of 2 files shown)


Talchain:
Merging over the same two pre-existing-class reds as #250, both verified NOT-this-diff: gates(windows-latest) colon-path (macos+ubuntu green) + audit fast-uri advisory GHSA-v2hh-gcrm-f6hx (reproduced on clean base; dedicated fix still outstanding).

PR #259, opened 2026-07-23
fix(v2): remove dead factor_evpi consumer (withhold EVPPI to S5) + OpenAPI↔runtime drift gate (Codex F3/F9)

Talchain:
Codex science-drive review **F3** (dead old-name consumer) and **F9** (OpenAPI↔runtime drift). Fable byte-review PASS.

**F3 — decision: WITHHOLD (justified at the bytes).** ISL #103 removed top-level `factor_evpi` and added outcome-unit `factor_evppi` + renamed `p_win_sensitivity`. The ranking-surface field the old consumer populated (`factor_sensitivity[].evpi_percentage_points`) is **win-probability percentage-points**; `factor_evppi` is **outcome-units**. Placing outcome units into a win-prob-pp field is a wrong-unit/wrong-label defect — the honest reconciliation + method-tagging is the S5 lane's job (D-23.8). So: **remove the dead consumer** (it could never fire against current ISL — PLoT was already on the self-disclosing heuristic for this surface), preserve the raw top-level passthrough of `factor_evppi`/`p_win_sensitivity` untouched, and do NOT fabricate a conversion. Corrects my earlier false "factor_evpi zero-consumers" claim.

**Cross-repo fail-loud guard (derive-don't-mirror, so a future ISL rename can't silently strand a consumer again):** a compile-time `isl-no-factor-evpi.type-pin.ts` (fails tsc if the field is re-declared on the ISL response type) + a runtime grep guard scanning `src/` for reads of the removed name, both with positive controls. Both were RED at base (flagged `v2-envelope.ts:151`), GREEN after removal.

**F9 — drift gate.** Single-source-of-truth `run-contract-keys.ts` (`V2_RUN_ALLOWED_KEYS` + enrichment keys + `islEnrichmentPassthrough()`); OpenAPI updated to document all accepted request keys + the four enrichment response fields + corrected stale `factor_evpi` wording; a `openapi-runtime-drift.test.ts` gate asserting the spec exactly equals the runtime allowlist/enrichment set (fail-loud, RED against the pre-F9 spec). The gate surfaced **four additional pre-existing undocumented request keys** (`goal_constraints`, `include_thresholds`, `include_e_values`, `include_voi`) beyond the named `factor_correlations` — all now documented for a clean set-equality (no exception list).

**Honest contradiction surfaced:** the byte-identity pin fixture is STALE (pre-#103, still carries `factor_evpi`), so removing the dead consumer moved that golden by exactly 2 fields on 2 non-lever factors (`below_resolution` → heuristic `evpi_percentage_points:0`). Against LIVE ISL the output is unchanged — the stale fixture merely stopped exercising a dead branch. Handled by deliberate regeneration + an explicit pin assertion (the dated capture was NOT falsified). structural-keys regenerated via the sanctioned tool (−6 keys, drift test green).

**Gates (baseline a07b0ae → branch):** `npm test` 5923→5921 pass / 0 fail (−7 dead-consumer −2 old-name-capture +3 F3 guard +1 pin +3 F9 gate = −2 net; +2 files) · typecheck 0 · openapi:lint unchanged · eslint 97→97 warnings (no headroom consumed). Expected CI red = the known windows colon-path.


Talchain:
Merging over 3 reds, all verified pre-existing-class, NONE introduced by this F3/F9 diff: (1) gates(windows-latest) = the documented colon-path standing red; (2) audit = a NEWLY-PUBLISHED find-my-way DDoS advisory (GHSA-c96f-x56v-gq3h) on a Fastify transitive dep — this diff touches ZERO dependencies, so it's pre-existing to the change (same class as the earlier fast-uri advisory #252); rowed for a dedicated deps PR; (3) validate-structure = entirely the pre-existing security-defined standing red (per-operation rule) — this diff adds only schema PROPERTIES to existing components, zero new operations, and produces ZERO non-security-defined errors, so the problem-set delta vs base is empty. The real gates (npm test 5921/0, typecheck 0, openapi:lint) are green.

PR #266, opened 2026-07-26
refactor(sequential-validation): one stage-definition predicate, normalised once [P1]

Talchain:
## P1 — one stage-definition predicate, normalised once

Follow-up hygiene on #265. #265's actual fix (making `validateSequentialGraph` total, killing the `TypeError: stage.decisions is not iterable` 500) was at the right altitude and is untouched here. This is about what ended up sitting beside it.

### The finding, verified at the bytes on `eceb6bb`

"Is this entry a stage definition?" was written out three times at three different strictnesses:

| site | object | `!Array` | numeric index | finite |
|---|---|---|---|---|
| `readStageDefinitions` (`:132`) | yes | yes | yes | yes |
| `buildNodeStageMap` (`:352`) | yes | **no** | yes | **no** |
| `getMaxStage` (`:400`) | **no** | **no** | yes | yes |

They diverged inside #265 itself, and the divergence is observable rather than cosmetic: an entry `readStageDefinitions` has already **rejected** as `INVALID_STAGE_DEFINITION` could still be consumed by the other two — seeding the node→stage map (fabricating a `FORWARD_REFERENCE` attributed to a discarded stage) and raising `getMaxStage`, which feeds the `maxStage > MAX_STAGES` 400 gate and `model_card.stages`.

### Scope honesty — this is latent, not a live 500

The two clauses that differed (the array check and the finite check) **cannot be tripped by a strict JSON request body**: JSON has no `Infinity`/`NaN` literal, and a JSON array cannot carry a named `index` property. For every JSON-representable input the three predicates agreed. So this is a latent-divergence fix — the next edit to any one site is what makes it bite. The tests construct the graph object directly, which is the honest level for the claim: this file's own header declares its input "UNTRUSTED JSON whatever the TypeScript types promise" and requires the function to be TOTAL, so its predicates are contract independent of any one caller.

### Change

- **`export function isStageDefinition(x: unknown): x is StageDefinition`** — the one predicate, used at all three sites.
- **`readStageDefinitions` now returns `NormalisedStage[]`** (`{index, label, decisions, resolvedUncertainties}`), resolving the id lists in the same pass; `buildNodeStageMap` consumes those.

That single move deletes:
- the optional `issues?` parameter on `readStageIdList` — whether a malformation got **reported** depended on which call site remembered to pass the sink;
- the second parse of the same stages (`:353-355`);
- the "omit `issues` on re-reads" convention a future editor had to remember (`:349-350`);
[...]

PR #267, opened 2026-07-26
fix(sequential): stop #265 blocking inputs that used to return 200 [C6]

Talchain:
## C6 — #265 blocks inputs that previously returned 200

#265's gate is **"any issue with `severity === 'error'`"**, which reaches the *pre-existing* codes (`NON_CONTIGUOUS_STAGES`, `MISSING_DECISION_NODE`, `MISSING_UNCERTAINTY_NODE`, `INVALID_NODE_STAGE`, `INVALID_DISCOUNT_FACTOR`), not only the four new malformation codes. Its own two new 400 tests exercise exactly one input — `stages = 'not-an-array'` — so the rest of the widened surface shipped uncovered.

### Measurement, not inference

Same probe, both routes, a **worktree per revision** — pre-#265 `04f6dbac` vs post-#265 `eceb6bb`:

| `sequential_metadata.default_discount_factor` | PRE | POST |
|---|---|---|
| `"0.95"` | **200**, `ev=6066` | **400** `INVALID_DISCOUNT_FACTOR` |
| `"abc"` | 200, `ev=null` | 400 `INVALID_DISCOUNT_FACTOR` |
| `true` | 200, `ev=6221.538` | 400 `INVALID_DISCOUNT_FACTOR` |
| `{}` | 200, `ev=null` | 400 `INVALID_DISCOUNT_FACTOR` |
| `null` | 200, `ev=6221.538` | 200 *(unchanged)* |
| `0.95` (control) | 200, `ev=6066` | 200, `ev=6066` *(unchanged)* |

The decisive row is the first. `"0.95"` returned **`ev=6066`, byte-equal to the numeric `0.95` control** — JS coerced the string in the `< 0 / > 1` comparisons and again in the arithmetic, so the client got the *right answer*. #265 turned that into a 400. The change is wider than the four input shapes above suggest: `/v1/analysis/sequential` was affected too, via its **pre-existing** gate at `:178`, not just the new policy-tree one.

### (a) Restore the numeric string — but disclose the coercion

A numeric string is a client-serialisation artefact, not a semantic error, so it is read as a number again and the request succeeds. The coercion is now **disclosed** as a `COERCED_DISCOUNT_FACTOR` warning instead of happening by accident. **Coercion is not permission:** `"1.5"` still fails exactly as `1.5` does.

`'abc'`, `true`, `{}`, `''`, `'  '` **stay error-severity on purpose.** Pre-#265 they returned 200 carrying `overall_expected_value: null`, or a boolean silently read as `1` — #265's block is *better* than that silence, so it is kept rather than restored, and that judgement is now written down in a test instead of implied. `coerceNumericString` is deliberately narrow for this reason: `Number('')` is `0` and `Number(true)` is `1`, and accepting either would invent a value the client never sent.

Also normalised at `analysis-sequential.ts:191`: the raw string previously reached `Math.pow`, the **ISL request body**, and `model_card.discount_factor` — whose declared type is `number` — so the response echoed `"0.95"` while the new warning said it had been read as `0.95`. A response must not say two different things about one value.

### (b) Correction — the reported mechanism does not hold

> *claimed:* a dropped stage entry cascades into `INVALID_NODE_STAGE`, "converting a previously-degraded-but-200 policy tree into a 400"

**Probed per input, that is not what happens:**

| `stages` entry | PRE-#265 | POST-#265 |
|---|---|---|
| `{label:'junk'}` (no index) | sequential **400** `NON_CONTIGUOUS_STAGES` / policy-tree 200 | 400 `INVALID_STAGE_DEFINITION` |
| `{index:'1'}` | sequential **400** `NON_CONTIGUOUS_STAGES` / policy-tree 200 | 400 `INVALID_STAGE_DEFINITION` |
| `42` / `null` / `[]` | **500** both | 400 `INVALID_STAGE_DEFINITION` |

1. On `/v1/analysis/sequential` there is **no 200 → 400** at all — every JSON-reachable bad entry already returned 400 (via `NON_CONTIGUOUS_STAGES`, because the un-filtered `stageIndices` contained a non-integer) or 500.
2. The 200 → 400 on `/v1/analysis/policy-tree` is caused by #265 **adding the gate at all** — the route previously discarded `validation` and returned 200 on *every* validation error — not by the cascade.
3. The cascade **changes no status**: `errors[0]`, the only issue any route reports, is always the `INVALID_STAGE_DEFINITION`, which is pushed first. Verified — one bad entry with three nodes at that stage yields `INVALID_STAGE_DEFINITION, INVALID_NODE_STAGE×3` with `errors[0].code === 'INVALID_STAGE_DEFINITION'`.

So the sub-claim that survives is the one about *quality*: those three `INVALID_NODE_STAGE` errors restate a consequence of the one real fault with the wrong subject, and `issues[]` **is** surfaced on the wire. Suppressed when the stage list is known-incomplete, **derived from the two lengths** rather than a counter that could drift. Not a weakening — the rejected definition is itself error-severity so the 400 stands, and a positive control proves `INVALID_NODE_STAGE` still fires whenever the stage list is intact.

**The real gap was cover for the policy-tree gate widening.** Three pre-existing error codes are now pinned on that route (`NON_CONTIGUOUS_STAGES`, `MISSING_DECISION_NODE`, `MISSING_UNCERTAINTY_NODE`) with a positive control that a well-formed graph is not blocked. That widening is *intended* — it matches the contract table this module has always published — it was simply untested.

### Evidence

**RED-first**, against unmodified `eceb6bb`: **4 failed | 11 passed**
```
× (a) RED-first: does not block either route          → expected 400 to be 200
× (a) RED-first: yields the SAME analysis as the equivalent number → expected 400 to be 200
× (a) RED-first: the coercion is DISCLOSED, not silent
× (b) RED-first: reports the rejected definition, not one error per orphaned node
      → expected [ 'INVALID_STAGE_DEFINITION', …(3) ] to not include 'INVALID_NODE_STAGE'
```
Every positive control passed in that same run — including *"the discount factor actually affects the result"* (`0.5` vs `1` give different `ev`), without which the same-answer equality could hold because the factor was ignored entirely, and *"`INVALID_NODE_STAGE` still fires when the stage list is intact"*.

**Mutation check** — throwaway worktree, uniquely-named path **outside** the repo root, each anchor asserted present before the write, worktree removed before the gate run. Each mutation kills only its own target:

| mutation | tests killed |
|---|---|
| A: remove the numeric-string coercion (restore #265's blanket block) | 3 |
| B: keep the coercion, drop the disclosure warning | 1 (disclosure) |
| C: un-suppress the `INVALID_NODE_STAGE` cascade | 1 (cascade) |
| D: revert the `model_card` numeric normalisation | 1 (same-answer) |

**Gate** — `npm test`, the repo's own gate:
- baseline `eceb6bb`: `562 files / 5953 passed, 29 skipped` → exit 0
- this branch, run 1: **1 failed** — `tests/stream.real.limited.test.ts`, `TypeError: fetch failed` / `ECONNRESET` on `${BASE}/stream`. Unrelated: that file has **zero** references to `sequential` or `discount`, and my diff touches neither streaming nor the server. Reported rather than hidden.
- this branch, run 2: **exit 0**, `563 files / 5968 passed, 29 skipped` → `+1 file / +15 tests`, exactly this PR's additions. Flake confirmed.
- `npm run typecheck` → exit 0. Pre-push validation (6 checks) PASSED. No test deleted, no baseline bumped.

### Sequencing

Branched from `staging` at `eceb6bb`, like the sibling PRs. Both this and #266 touch `src/util/sequential-validation.ts`; the hunks are in different regions (this one: the discount-factor block and the node-stage loop; #266: the stage predicate and `buildNodeStageMap`) so any conflict is textual and small. **Merge this one first** — it is a live behaviour change, the others are hygiene.

**Do not merge without review** — `staging` is unprotected (protection API returns 404), so the merge *is* the gate.

### Calibration

Still a dormant cluster: `validateSequentialGraph` has four callers, all `/v1/analysis/*` + `/v1/explain/policy`, and `/v1/run` (what CEE actually calls) never imports it. So this restores correctness for a client that does not exist yet — worth doing because the regression was undisclosed, not because anyone is hitting it.

### DIFF
--- a/src/routes/v1/analysis-sequential.ts
+++ b/src/routes/v1/analysis-sequential.ts
@@ -187,11 +187,23 @@ export async function registerSequentialAnalysisRoute(app: FastifyInstance) {
         }
       }
 
-      // Get discount factor
-      const discountFactor =
+      // Get discount factor.
+      //
+      // Read as a number even when the client sent a numeric string. Without
+      // this the raw string reached `Math.pow` (which coerced it silently), the
+      // ISL request body, and `model_card.discount_factor` — whose declared type
+      // is `number` — so the response echoed `"0.95"` while the accompanying
+      // COERCED_DISCOUNT_FACTOR warning said it had been read as `0.95`. The
+      // response must not say two different things about one value.
+      // `validateSequentialGraph` has already blocked a non-coercible
+      // `sequential_metadata.default_discount_factor`; the range check below
+      // still guards the top-level `discount_factor`, which nothing validates.
+      const rawDiscountFactor =
         body.discount_factor ??
         body.graph.sequential_metadata?.default_discount_factor ??
         1.0;
+      const discountFactor =
+        typeof rawDiscountFactor === 'number' ? rawDiscountFactor : Number(rawDiscountFactor);
 
       if (discountFactor < 0 || discountFactor > 1) {
         return replyWithAppError(reply, {
--- a/src/util/sequential-validation.ts
+++ b/src/util/sequential-validation.ts
@@ -32,6 +32,17 @@
  * - `INVALID_STAGE_DEFINITION` (error): A `stages[]` entry is not a stage object
  * - `MISSING_STAGE_ID_LIST` (warning): Stage omits `decisions`/`resolved_uncertainties`
  * - `INVALID_STAGE_ID_LIST` (warning): Stage id list present but not an array
+ * - `COERCED_DISCOUNT_FACTOR` (warning): Discount factor sent as a numeric string
+ *
+ * ## Blocking is a wire contract — check the pre-#265 behaviour before widening it
+ *
+ * All four routes block on ANY error-severity issue, so raising a code from
+ * warning to error, or adding an error for a shape that previously passed, turns
+ * a 200 into a 400 for every client sending that shape. #265 did this to
+ * `INVALID_DISCOUNT_FACTOR` (a JSON-string discount factor had returned 200 with
+ * the CORRECT result, coerced by JS) and, by adding the policy-tree gate at all,
+ * to every pre-existing error code on that route. Neither was covered by a test.
+ * Before changing a severity here, measure the previous behaviour at the route.
  *
  * ## This function must be TOTAL
  *
@@ -69,6 +80,22 @@ export interface SequentialValidationResult {
 /** Total array read — untrusted input may put anything here. */
 function asArray<T>(value: unknown): T[] {
   return Array.isArray(value) ? (value as T[]) : [];
+}
+
+/**
+ * A numeric string read as a number, else `undefined`.
+ *
+ * Deliberately narrow: only a string whose whole trimmed content is a finite
+ * numeric literal. `''`, `'  '`, `'abc'`, booleans, objects and arrays are NOT
+ * coercible — `Number('')` is 0 and `Number(true)` is 1, and accepting either
+ * would invent a value the client never sent.
+ */
+function coerceNumericString(value: unknown): number | undefined {
+  if (typeof value !== 'string') return undefined;
+  const trimmed = value.trim();
+  if (trimmed === '') return undefined;
+  const parsed = Number(trimmed);
+  return Number.isFinite(parsed) ? parsed : undefined;
 }
 
 /** Human-readable stage reference for issue messages, safe on a missing label. */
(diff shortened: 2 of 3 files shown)


Talchain:
Closing as moot: #268 (merged) deleted the sequential routes whose validation behaviour this PR fixed, per the 26 Jul Paul-ratified vacuity ruling. The C6 defect class (new 400s on previously-200 inputs) died with the routes. #265's general unhandled-error logging was preserved in #268.

PR #287, opened 2026-07-27
feat(v2/run): emit driver_order — one canonical driver ordering + attestation (family-4 S1)

Talchain:
## Family 4 (driver rankings) — slice **S1**, the PLoT producer half

PLoT now publishes **one canonical driver ordering and attests how it made it**, as an
additive top-level `/v2/run` field. This is the producer half of the ratified authority
model — **ISL measures · PLoT orders + attests · CEE permits + projects · UI renders
without reordering** (`MEANING-LAYER-FAMILY4-AUTHORITY-AMENDMENT-2026-07-27.md` §2, §4.3,
§6, §8-S1).

### ⚠ Scope: ADDITIVE ONLY — and this is narrower than the pack's S1

The design pack's S1 has two halves. This PR ships the **first** half only:

| Half | Status here |
|---|---|
| Emit `driver_order` (§4.3) with the tie verdict (§6) | ✅ **this PR** |
| Make the five #1-naming surfaces PROJECTIONS of `ranked_factor_ids[0]` | ⛔ **NOT here** — deferred, enumerated in *Residual* below |

Nothing that ranks, crowns, or bands today changed shape, value, or meaning. The new order
is emitted **alongside** the existing divergent ones, not swapped in — the staged-rollout
doctrine (older consumers drop unknown fields silently), and §4.4's deploy-order constraint
(*"the safe order is consumer-first"*): the lever demotion is today the only thing keeping
`importance_rank == 1` off a factor the producer explicitly zeroed, so it is **attested,
not removed**.

**Additivity is proven, not asserted** — the regenerated golden's complete diff is:

```
top-level keys ADDED:   ['driver_order']
top-level keys REMOVED: []
top-level keys CHANGED: ['_meta']  ->  only _meta.response_content_hash
response_hash:          60e3ac213554be4f  (UNMOVED — it canonicalises the REQUEST)
```

### What S1 emits

`driver_order`, a top-level sibling of `factor_sensitivity[]`:

```ts
driver_order: {
  version: 1,
  basis: 'graph_structural' | 'isl_uncertainty' | 'none',
  ranked_factor_ids: string[],          // THE canonical order — ids only
  species: 'single' | 'mixed_graph_isl',
  lever_policy: 'du_union' | 'stamp_only' | 'none',
  lever_ids: string[],                  // levers are MARKED, not hidden
  separability:   { top_pair_separable: boolean | null, method: string | null },
  rank_stability: { max_rank_flip_rate: number | null, min_attribution_stability: 'negligible'|'low'|'moderate'|'high' | null },
}
```

**The ordering rule, stated at the bytes** (`src/lib/driver-order.ts` header, and
re-derivable from the payload + the request's option interventions):

1. **Basis order** — `computeFactorSensitivityFromGraph`'s influence order on the
   graph-primary path (`basis: 'graph_structural'`); ISL's own Monte-Carlo
   uncertainty-importance order on the ISL-only fallback (`isl_uncertainty`).
2. **ISL-only tail** — `mergeIslConfidenceIntoGraphFactors` appends non-lever ISL-only rows
   at the tail **with no re-sort**, so within-partition order is the producer's *assembly*
   order, not a global sort on one number. That state is what `species: 'mixed_graph_isl'`
   exists to disclose; before this field no consumer could detect it at all.
3. **Lever partition** — `applyLeverAwareImportanceOrder`'s stable partition, every
   non-lever ahead of every lever, D-U union identity (ISL's `intervention_override` stamp
   **OR** membership of `interventionTargetIdsFromOptions(options)`) ⇒ `lever_policy: 'du_union'`.

**Absence semantics.** Emitted whenever `factor_sensitivity` is emitted — *including when
that array is empty*, as `basis: 'none'` rather than omission. This closes the ambiguity the
per-row `importance_basis` still has (its emission is inside the ISL branch, so absence
could mean old payload, non-ISL branch, or dropped key). Same present-empty-vs-absent
resolution #284 already paid for on the family-2 side, adopted rather than re-litigated.

**ISL's `suppressed_attributions` now has a reader.** PLoT had zero references to it. It is
carried into `basis: 'none'` on the path where the suppressed producer *was* the basis, so
`not_computed` is **sourced from the producer that knows** instead of inferred from an empty
array. It deliberately does **not** blank the basis on the graph path — calling a real
graph-derived order `'none'` would be a different lie.

**The tie verdict — and the one thing this build refuses to say.** `top_pair_separable: true`
is **never emitted**. Deciding *separable* from ISL's bootstrap statistics needs a
**threshold**, and no threshold for the DRIVER order is ratified; T3 requires the producer to
publish the one threshold it used, and inventing one here would recreate exactly the
three-thresholds-in-three-repos defect T3 exists to kill (PLoT 0.10 / CEE 1.0 / UI 0.65 on
the *options* question). So the verdict is monotone-honest:

- `false` + `method: 'basis_value_exact_tie'` — the top two rows are **exactly equal** on the
  basis quantity. A proven non-separation; no threshold involved.
- `null` + `method: null` — **UNRESOLVED**. Per T2 every consumer fails closed; strict
  inequality is *not* evidence of statistical separability.

`rank_stability` carries the *measurements* a future verdict would read (worst
`rank_flip_rate`, worst `attribution_stability` band) — no threshold applied, `null` never
coalesced to 0.

⛔ **Open decision this PR does not take, flagged for S4:** which statistic and which
threshold decide `top_pair_separable: true`, against which pair. Until it is ratified, CEE's
`separability`-fed permission will resolve to fail-closed on almost every payload. That is
the honest state of the producer, not a defect introduced here — but it is a real product
consequence and should be sequenced deliberately.

### Evidence

**RED-first** — the new specs applied to the pre-S1 tip `a5ffa60b` in a throwaway worktree
**outside the repo root**: **13 of 19** fixture assertions RED, the unit spec RED at import.
The 6 green ones are the positive controls and the additivity pins — correct, they describe
the world S1 does not change.

**Self-consistency, not a mirror.** `tests/driver-order-attestation.fixture.test.ts`
deliberately does **not** import `src/lib/driver-order.ts`: it re-derives lever identity from
the **request**, re-applies the declared ordering rule itself, and compares. It also carries
a control proving the rule is *load-bearing on this fixture* — the raw influence order is a
**different** order (a lever tops it), so the assertions cannot pass for the wrong reason.

**Mutation-checked** (throwaway worktree outside the repo root, removed before any gate run):

| Mutant | Result |
|---|---|
| M1 — ordering rule perturbed (reverse `ranked_factor_ids`) | **5 RED** (parallel-to-array, declared-rule re-derivation, lever partition, lever_ids, brief parity) |
| M2 — `basis` graph↔isl swapped | **5 RED** |
| M3 — fabricate `top_pair_separable: true` on strict inequality | **2 RED** |
| M4 — `rank_stability` reports best instead of worst band | **2 RED** |
| M5 — lever marking downgraded to the stamp-only predicate | **4 RED** (unit spec only — *both* golden levers are ISL-stamped, so the fixture cannot see this; recorded, not hidden) |
| M6 — conditional emission restored (omit on `basis: 'none'`) | **2 RED** |

**Gates** (run at this tip, from `package.json`, not from memory): `npm run typecheck` ✅ ·
`npm run build` ✅ · `npm run validate:contracts` ✅ · `eslint --max-warnings 97` ✅ (86
warnings, 0 errors) · `node tools/validate-openapi-structure.js` ✅ · full `npm test` — see
the checklist comment.

**Spec-touching CI disclosure.** `openapi-sanity / validate-structure` runs on this PR and
its Redocly step **will be red — it is red on `staging` too**. Proven empty base-vs-branch
delta, same conditions, same working tree, ignore-file applied:

```
base (a5ffa60b contracts/openapi.yaml):  37 errors, 28 warnings
this branch:                             37 errors, 28 warnings
```

Spectral (`npm run openapi:lint`) and the structure validator both pass. Branch protection
re-derived at push time: `gh api .../branches/staging/protection` → **HTTP 404, not
protected** — the push is the gate.

### Residual — the divergent ordering sources S1 does **not** touch

Enumerated so none is inherited as fixed, and pinned by
`tests/driver-order-attestation.fixture.test.ts` so none can drift silently before the slice
that reconciles them. Measured on the committed 2026-07-07 golden:

| Surface | Ranks on | Lever-aware? | Agrees with `ranked_factor_ids[0]`? |
|---|---|---|---|
| `factor_sensitivity[].importance_rank` | this order | ✅ D-U union | ✅ it **is** this order |
| `factor_sensitivity[].driver_label === 'biggest'` | argmax `influence_score` (`lib/driver-label.ts`) | ❌ | ❌ **crowns the option-pinned lever** (`sensitivity_score: 0`, `elasticity: 0`) |
| `m1_coaching.key_drivers[].rank` | `abs(influence_score ?? elasticity ?? 0)` (`coaching/key-drivers.ts`) | ❌ | ❌ names the same lever |
| `dominant_factor` | `detectDominantFactor` over unfiltered rows (`trust/factor-dominance.ts`) | ❌ | suppressed here only by its `>2` ratio gate |
| `decision_brief.top_drivers[0]` | `filterInterventionOverrides` = ISL **stamp only** (`assembly/decision-brief.ts`) | ⚠ stamp-only | ✅ on this fixture — the stamp happens to cover |
| facts-path `importance_rank` | positional `idx + 1` (`routes/v2/run.ts`, `facts/mapper.ts`) | n/a — a **position**, not a rank | mirrors the array, so agrees by accident |

Also untouched, and untouched deliberately: the lever **demotion** itself (§4.4 — un-demoting
before CEE permits and the UI consumes would re-arm the family's worst live defect on the one
surface that is currently correct); `robustness.near_tie`'s 0.10 threshold; the CEE-side and
UI-side thresholds.

### Notes for the reviewer

- **Per-row `importance_basis` is now a mirror, and it fails loud.** §4.3's transition rule:
  keep it for one release with a test asserting **every row agrees with `driver_order.basis`**;
  delete it in the release after. That assertion is in the fixture spec.
- **`tools/gen-structural-keys.mjs` gained `src/lib/driver-order.ts` as a contract file** —
  the object's members are declared there, not in `engine-v3.ts`, so without it the
  attestation's keys would be digested out of redacted log payloads. Derived, not hand-listed;
  the drift test enforces it. It also contributes four *input-only* property names that can
  never occur as payload keys — a documented no-op, chosen over hand-filtering the list.
- **`§2.2` vs `§4.3` discrepancy in the pack, resolved toward §4.3 (non-blocking).** The §2.2
  seam sketch names an attestation member `suppressed_ids`; the §4.3 typed object — which
  S1's own bullet cites — has `lever_ids` and no `suppressed_ids`. ISL's
  `suppressed_attributions` members are *attribution names* (`'factor_sensitivity'`,
  `'p_win_sensitivity'`), **not factor ids**, so nothing in the estate can currently source a
  `suppressed_ids` list. Implemented §4.3 exactly; flagging rather than improvising a member
  whose semantics are not determined.


Talchain:
### Gate results at head `287ec06`

| Check | Result |
|---|---|
| `test` | ✅ pass |
| `gates (ubuntu / macos / windows)` | ✅ pass — the historical windows colon-path flake did **not** fire |
| `audit` | ✅ pass — green, as expected; a red here would be real |
| `guard` · `verify` · `safety` ×2 · `smoke` ×2 · `update_release_draft` | ✅ pass |
| `openapi-sanity / validate-structure` | ❌ **fail — pre-existing, empty delta (proof below)** |

**Local gates at this tip** (commands read from `package.json`, not from memory):
`npm run typecheck` ✅ · `npm run build` ✅ · `npm run validate:contracts` ✅ ·
`eslint src tests --max-warnings 97` ✅ (86 warnings, 0 errors) ·
`node tools/validate-openapi-structure.js` ✅ · `npm run openapi:lint` (Spectral) ✅ (0 errors, 4 warnings) ·
full `npm test` ✅ **578 files / 6250 tests passed, 0 failed** · pre-push hook ✅ **6/6 checks, 6256 tests**.

#### `validate-structure` — read at the job output, not from a registry

The failing step is **"Validate OpenAPI schema (Redocly)"**:

```
❌ Validation failed with 37 errors and 28 warnings.
##[error]Process completed with exit code 1.
```

Base-vs-branch delta, measured **same conditions, same working tree, ignore-file applied**:

| `contracts/openapi.yaml` at | Redocly result |
|---|---|
| `a5ffa60b` (pre-PR staging) | **37 errors, 28 warnings** |
| this branch | **37 errors, 28 warnings** |

⇒ **Zero delta.** This job is a standing red on `staging` that only becomes visible on a
spec-touching PR (it is `paths:`-filtered to `contracts/openapi.yaml`). The two steps this
change could plausibly have broken both pass: the structure validator ran before Redocly and
passed in CI, and Spectral passes locally with 0 errors.

⚠ A red everyone learns to ignore is a broken alarm. Worth a separate row: either fix the 37
or generate the ignore file, so the next spec PR gets a signal instead of noise.

#### Flake note, disclosed

The **first** pre-push run failed on `tests/rate-limit.ipv6.test.ts` (2 assertions, 429
expected / 200 received). Not this change: it passes in isolation, and a same-conditions
`npm test` in a throwaway worktree at `origin/staging` (`dfeef56`) was fully green
(577 files). The retry passed the hook 6/6. It is timing-sensitive under a loaded 583-file
parallel run — a candidate for its own hardening row.

PR #288, opened 2026-07-28
feat(v2/run): five #1-surfaces project driver_order[0] + provisional separability default (family-4 S1b)

Talchain:
**Family 4, slice S1b — the consumer half of "one order"** (amendment §8-S1 remainder, ROADMAP 1.344), plus the Paul-ratified **provisional separability default** (28 Jul).

S1 (#287) made PLoT publish exactly one canonical driver ordering. It changed none of the surfaces that crown, and on the committed golden two of them named the **option-pinned lever the same response publishes at `sensitivity_score: 0`, `elasticity: 0`, `zero_reason: 'intervention_override'`**. This PR makes all five #1-naming surfaces projections of `ranked_factor_ids[0]`, and gives `separability` a provisional verdict instead of a permanent `null`.

**Advances:** finish-line criterion 7 (differentiated science surface) and criterion 1 (family-4 drivers) — the canonical driver order is now what a user actually sees.

---

### The statistic, and its provenance

| | |
|---|---|
| **Statistic** | relative gap `(first − second) / first` on the basis quantity (`influence_score`) across the top pair of the canonical order |
| **Threshold** | `0.10` — **bound to `NEAR_TIE_THRESHOLD`** (`src/trust/result-coherence.ts:25`), the repo's one ratified near-tie magnitude, already used by `computeNearTie` and by the brief's `'very_close'` band |
| **On the wire** | `method: 'relative_gap_0.10_provisional'` — derived from the constant via `.toFixed(2)`, so the string cannot drift from the number it describes |
| **Ratification** | 🟡 PROVISIONAL. Neil owns the replacement; the one-line overrule is documented at the function |

The number is **bound, not copied** (trap 12), and the binding is *also* pinned to the literal `0.10` — so if the options-side constant is ever changed for an unrelated product reason that is a LOUD failure here rather than a silent move of the driver verdict.

### ⚠ Refuting the brief: the existing convention is ABSOLUTE, not relative

The brief said *"a 0.10 **relative-gap** threshold exists in the codebase — find it, cite it."* **It does not.** `NEAR_TIE_THRESHOLD` is applied as an **absolute** gap between two `win_probability` values (`src/routes/v2/run.ts`: `const gap = topWinProb - secondWinProb; const isTie = gap < NEAR_TIE_THRESHOLD`), and its own comment says *"gap between top two options < 10%"*. There is no relative-gap convention in this repo.

So I reused the **number** and deliberately changed the **form**, for a reason specific to this quantity rather than to taste:

`influence_score` is `|influence| / maxAbsInfluence` (`src/lib/factor-influence.ts:547`) — **normalised by the largest row, which after the lever demotion need not be in the compared pair at all.** On the committed golden the max row *is* the demoted lever (`influence_score: 1`), while the two rows actually compared sit at `0.497` and `0.390`. Had that lever's raw influence been twice as large, both would halve to `0.249 / 0.195`: **same underlying data, absolute gap collapsing 0.107 → 0.053, verdict flipping** on a number outside the comparison. The relative gap is invariant to that rescaling (0.2145 either way). `win_probability` has no such problem — it is normalised across the options being compared — which is why the same constant is right and the same form is not.

### The requirements, discharged

| Req | Where |
|---|---|
| (a) labelled PROVISIONAL, names the pending Neil ratification + the one-line overrule | `src/lib/driver-order.ts` — the `DriverOrderSeparability` doc and the `PROVISIONAL_…` const |
| (b) `method` states statistic + threshold + provenance | `relative_gap_0.10_provisional`, derived from the constant |
| (c) exact-tie path unchanged | checked **first**, keeps `basis_value_exact_tie`; mutant **M12** proves relabelling it REDs |
| (d) fail-honest preserved | absent / null / non-finite ⇒ `null`; mutant **M13** proves a `?? 0` REDs |
| (e) partition-straddle LOW | **addressed** — see below |

### (e) The partition-straddle edge — addressed, not accepted

S1 compared rows 0/1 of the **final** order, i.e. after `applyLeverAwareImportanceOrder` partitions levers to the back. Those rows are not necessarily adjacent in any sort on `influence_score`:

- **Across the lever partition** — a lever keeps its true structural influence (only sensitivity/elasticity/VOI are zeroed) and is frequently the largest. With one non-lever and one lever, `b > a` and the gap is meaningless. Worse, an **exact equality across the partition would have published `basis_value_exact_tie` — "proven not separable" — for a coincidence between two rows the producer never compared.**
- **Across two species** — in `mixed_graph_isl`, graph rows carry path-analysis influence and appended ISL rows carry Monte-Carlo uncertainty importance, under the same field name.

Both now fail **closed to `null`**, deliberately not to `false`: `false` is a positive claim the producer has not earned. The guard can only ever *remove* a verdict. A residual is accepted knowingly and stated at the function: the guard proves the pair is **comparable**, not that the estimates are statistically distinguishable — that is Neil's question, and it is why `method` says `provisional`.

### Also disclosed rather than papered over

Near its own threshold the verdict is a floating-point coin-toss — `(1 − 0.9)/1 = 0.09999999999999998` lands NOT separable while `(0.4 − 0.36)/0.4 = 0.10000000000000009` lands separable. Pinned as a **disclosed limitation** (no epsilon added): it is what a gap between two point estimates is worth at the last bit, and one more argument for the provisional label.

---

## Part 2 — the five #1-surfaces become projections

| # | Surface | Before | After |
|---|---|---|---|
| 1 | `driver_label === 'biggest'` | argmax `influence_score`, lever-blind — **crowned the lever** | `ranked_factor_ids[0]` |
| 2 | `m1_coaching.key_drivers[0]` | `abs(influence_score ?? elasticity ?? 0)` sort — **crowned the lever** | `importance_rank` ascending (stable) |
| 3 | `dominant_factor` | internal argmax over unfiltered rows | `factors[0]` only; **gates unchanged** |
| 4 | `decision_brief.top_drivers[0]` | stamp-only lever predicate + a second `\|elasticity\|` sort | canonical order minus `driver_order.lever_ids` (D-U union) |
| 5 | facts-path `importance_rank` | positional `idx + 1` | `fs.importance_rank` (extracted as `mapFactorSensitivityToFactsInput`, so the claim is testable) |

Two details worth review attention:

- **`dominant_factor`'s rival is the strongest row anywhere in the order, not `factors[1]`.** The canonical order demotes levers while they keep their true influence, so the second-largest number is often not in second position. Comparing rank 1 only to rank 2 would call a 0.6 factor "dominant" while a 1.0 factor sat three rows down — *a new fabrication introduced by the fix*. Pinned.
- **`top_drivers` fails CLOSED on absence.** With no `driver_order` the brief keeps exactly its pre-S1b behaviour; absence of an attestation is not permission to publish a lever.

### ⛔ §4.4 honoured — the lever demotion stays

`applyLeverAwareImportanceOrder` is untouched. Levers are still demoted and MARKED, never un-demoted; the raw structural argmax is still published under its own honest name as `influence_rank === 1`. Pinned by two assertions in the projection spec.

### This supersedes a deliberate prior ruling, on re-derived grounds

`importance-authority.ts:117-140` left `driver_label` divergent on three reasons. All three are addressed in `src/lib/driver-label.ts` rather than waved past — and note the third, *"blast radius zero — no consumer reads `driver_label`"*, was a census taken 25 Jul at tips that have since moved. **Not inherited** (trap 12); and the argument cuts the other way regardless — a field with no readers is the cheapest possible moment to correct it.

---

## ⭐ The flip enumeration — every golden change, and why each is the canonical order asserting itself

The complete diff of `tests/fixtures/isl-v2-live-20260707/plot-v2-run.golden.json` is **five hunks and nothing else**:

| # | Change | Verdict |
|---|---|---|
| 1 | `fac_hiring_cost.driver_label`: `moderate` → **`biggest`** | ✅ The canonical #1 takes the crown. |
| 2 | `fac_tech_lead.driver_label`: `biggest` → **`strong`** | ✅ **The intended product fix.** The lever loses the crown and is labelled by its own number (`influence_score: 1` ⇒ `strong`, the unchanged magnitude band). This factor publishes `sensitivity_score: 0`, `elasticity: 0`, `zero_reason: 'intervention_override'` in the same response — it was being displayed as the biggest driver in the decision. |
| 3 | `driver_order.separability`: `{null, null}` → **`{true, relative_gap_0.10_provisional}`** | ✅ Part 1. Re-derived in the pin from the golden's own rows: `(0.4971 − 0.3905)/0.4971 = 0.2145 ≥ 0.10`. |
| 4 | `m1_coaching.key_drivers` reordered `[tech_lead, dev_headcount, hiring_cost, team_maturity]` → **`[hiring_cost, team_maturity, tech_lead, dev_headcount]`**, ranks reassigned 1–4 | ✅ Exactly `ranked_factor_ids`. Both levers move to the back; no row added, dropped or revalued. |
| 5 | `_meta.response_content_hash` moves | ✅ Correct — it hashes the public semantic surface, and the surface changed. Re-pinned. |

**What did NOT move, and was checked:**

- `response_hash` — **unmoved** (`60e3ac213554be4f`). It canonicalises the *request*; a crown change must not move the UI's freshness token.
- `factor_sensitivity[]` — array order, `importance_rank`, `influence_rank`, and every numeric value **unchanged**.
- `decision_brief.top_drivers` — **unchanged in value.** The stamp happened to cover on this capture; only the derivation changed. The separating input (an **unstamped** D-U lever, which the stamp misses) is a unit leg.
- `fact_objects` — **unchanged in value.** Position and rank coincide on every live payload by Rule S3; only the provenance changed.
- `dominant_factor` — **still absent.** But the *suppressing gate moved*, and that is itself the fix: the candidate is now the canonical #1, which fails the `> 0.5` influence floor, where before the candidate was the lever, suppressed only by the ratio gate at `1 / 0.7243 = 1.38`. The F-D3 leg opens that ratio gate and proves the lever still cannot be crowned.

---

## RED-first evidence (at the pre-fix bytes)

| Spec | RED |
|---|---|
| `driver-order-projection.fixture.test.ts` | **4 failed / 9 passed** — `driver_label` and `key_drivers[0]` both `fac_tech_lead`, expected `fac_hiring_cost` |
| `driver-order-separability.unit.test.ts` | **12 failed** — constants absent, every provisional branch `null` |
| `driver-surface-projection.unit.test.ts` | **8 failed** — F-D3 crowned `fac_tech_lead`; `top_drivers[0]` was the unstamped lever `Salary Cost`; the facts mapper did not exist |

The three surfaces that already agreed on the golden are covered at the units, because **a fixture that agrees by coincidence cannot RED on the defect** — each leg builds the one separating input that breaks the coincidence, all sourced from the real capture rather than hand-invented.

## Mutants — every one bites (throwaway worktree OUTSIDE the repo root, removed before gates)

| Mutant | Result |
|---|---|
| M1 separability default removed | 12 failed |
| M2 comparability guard removed | 2 failed |
| M3 threshold drifted 0.10 → 0.20 | 5 failed |
| M4 `driver_label` crown reverted to argmax | 12 failed |
| M5 `key_drivers` reverted to impact sort | 6 failed |
| M6b `factor-dominance.ts` reverted to **pre-S1b bytes** | 3 failed |
| M7 `top_drivers` reverted to stamp-only | 3 failed |
| M8 facts rank reverted to positional | 1 failed |
| M9 new unregistered numeric quantity added | 1 failed |
| M10 register entry made stale (field renamed) | 2 failed |
| M11 derivation stops recursing into nested literals | 2 failed |
| M12 exact-tie relabelled provisional | 3 failed |
| M13 fail-honest broken (`?? 0`) | 1 failed |

**Reported honestly:** my first M6 was **malformed** — I restored the internal sort but left `factors.slice(1)`, so `top1` fell inside its own rival set and the gate self-suppressed. It showed 0 failures. That was a bad mutant, not a weak test; **M6b** reverts the whole file to its pre-S1b bytes and bites.

## §3.2 — the derived producer register

`tools/derive-driver-quantities.mjs` extracts, **by AST**, every numeric field on `FactorSensitivityResultV3` and `EdgeSensitivityResultV3` (14, including one nested a level down inside an inline object literal). `src/lib/driver-quantity-register.ts` declares `{role, disposition, unit, sign, note}` for each. The gate fails in **both** directions — an unregistered quantity and a stale entry — and **UNPARSEABLE fails loud**: a missing file or renamed type *throws*, never returns an empty domain that would make the gate pass by testing nothing. Same instrument and same `typescript` dependency as the existing `tools/gen-structural-keys.mjs`. Exemptions are **emitted as data** (empty today, and the emptiness is the claim).

## The other two S1-review LOWs, ridden

- **ISL-only-path `lever_policy` wording** — `lever_policy` names the **predicate**, not the treatment, and the treatment differs by path: levers are *demoted* on the graph path and *removed upstream* on the ISL-only fallback. Now stated, with the consequence spelled out: **`lever_ids: []` means "no lever is IN this order", never "this decision has no levers".**
- **The latent `factorSensitivitySource ?? 'isl'` default** — `SensitivityData.factorSensitivitySource` and `.structuralLeverIds` are now **required**, so the only place the value can be absent is the genuine raw-ISL fallback branch (where `'isl'` is a derived fact, not a default). A future path omitting it is now a compile error instead of a silently wrong `basis`.

---

## Gates

| Gate | Result |
|---|---|
| `bash scripts/pre-push-validate.sh` | ✅ **PASS 6/6** |
| `npm test` | ✅ **6318 passed, 0 failed** (583 files) |
| `npm run typecheck` | ✅ clean |
| `npm run spec:lint` | ✅ 0 errors (4 pre-existing warnings) |
| `node tools/gen-structural-keys.mjs --check` | ✅ up to date |
| **Baseline `36b3309` in a separate clone** | ✅ 6256 passed, 0 failed — so every delta is mine |

### Known reds, re-derived at these bytes rather than cited

- **`validate-structure` — RED, and pre-existing.** This PR touches `contracts/openapi.yaml` so the job will run. Re-proven against the pristine baseline: `tools/validate-openapi-structure.js` **passes on both**; the red is the Redocly step, at **37 errors / 28 warnings on baseline and 37 errors / 28 warnings on this branch**, with an **empty delta on the finding set**.
- **`npm run lint` — 86 warnings, 0 errors, on baseline AND branch.** Unchanged; nothing absorbed, nothing added.
- **Branch protection re-derived at push time:** `HTTP 404 "Branch not protected"` ⇒ `staging` is unprotected, so the push is the gate.
- One flake observed and dismissed with evidence: `tests/rate-limit.ipv6.test.ts` timed out in a server-spawn hook on one pre-push run; it passes in isolation on **both** this branch and the pristine baseline, and the re-run was clean.

---

## What a reviewer should attack

1. **The chosen form (relative, not absolute).** The max-normalisation argument is the whole case; if `influence_score`'s normaliser is ever changed, this reasoning needs redoing.
2. **`dominant_factor`'s strongest-rival comparison** — the one place the fix could have introduced a *new* fabrication, and the reason it is not `factors[1]`.
3. **Crowning rank 1 when its `influence_score` is absent.** Deliberate: `'biggest'` is now a rank claim, and the producer *has* ranked those rows. The row still carries no magnitude band. Pinned, and flagged as a behaviour change.
4. **Whether `top_drivers` should have kept the `|elasticity|` sort within non-levers.** It is value-identical on the golden; I dropped the second sort because "PLoT owns exactly one ordering" is the point of the slice.

**Do not merge without review.** Deploys to staging on merge.

### DIFF
--- a/contracts/openapi.yaml
+++ b/contracts/openapi.yaml
@@ -6242,12 +6242,22 @@ components:
             producer, or a dropped key) and a consumer must fail closed on it.
 
 
-            ADDITIVE. `factor_sensitivity[].driver_label`, `dominant_factor`,
-            `m1_coaching.key_drivers[].rank`, `decision_brief.top_drivers[0]`
-            and the facts-path `importance_rank` are UNCHANGED by this slice
-            and remain independent argmaxes; on the pinned 2026-07-07 capture
-            two of them still name a different factor from
-            `ranked_factor_ids[0]`. Reconciling them is a later slice.
+            ⭐ EVERY #1-NAMING SURFACE IN THIS RESPONSE PROJECTS
+            `ranked_factor_ids[0]` (family-4 S1b): `factor_sensitivity[].driver_label
+            === 'biggest'`, `dominant_factor`, `m1_coaching.key_drivers[0]`,
+            `decision_brief.top_drivers[0]` and the facts-path
+            `importance_rank`. They are no longer independent argmaxes, so a
+            consumer reading any of them gets the same answer as reading this
+            field. Before S1b two of them named the option-pinned lever the
+            same response publishes at `sensitivity_score: 0`.
+
+
+            ⚠ Option-controlled levers are still DEMOTED to the back of this
+            order and MARKED in `lever_ids`, not removed — ranking a lever in
+            its true place is the honest end state but waits on CEE's
+            permission layer. The RAW structural influence order is still
+            published under its own name as
+            `factor_sensitivity[].influence_rank`.
 
 
             Not in `response_hash` (which canonicalises the REQUEST).
@@ -6329,21 +6339,47 @@ components:
                   type: boolean
                   nullable: true
                   description: >-
-                    `false` — PROVEN not separable (the top two rows are
-                    EXACTLY equal on the quantity the order was made on; no
-                    threshold involved). `null` — UNRESOLVED; a consumer MUST
-                    fail closed, and strict inequality on the basis quantity is
-                    NOT evidence of statistical separability. `true` is NOT
-                    emitted by this build: no threshold for the DRIVER order is
-                    ratified, and inventing one would recreate the
-                    three-thresholds-in-three-repos defect the option-side
-                    `near_tie` already suffers.
+                    🟡 PROVISIONAL (ratified 2026-07-28; the statistic is
+                    pending science sign-off — ALWAYS read `method`).
+                    `false` — not separable. `true` — separable. `null` —
+                    UNRESOLVED, and a consumer MUST fail closed on it: absence
+                    of a verdict is never permission.
+
+
+                    `null` is returned whenever the producer cannot honestly
+                    decide: fewer than two rows; either basis value absent or
+                    non-finite (never coalesced to 0); the top two rows on
+                    OPPOSITE sides of the lever partition, or of two different
+                    row species — in those states the emitted order is not a
+                    sort on the compared quantity, so any verdict about the
+                    pair would be a fabrication.
                 method:
                   type: string
                   nullable: true
                   description: >-
-                    Names the method behind a non-null verdict
-                    (`basis_value_exact_tie`); `null` when unresolved.
+                    Names the STATISTIC, THRESHOLD and RATIFICATION STATUS
+                    behind a non-null verdict; `null` when unresolved. One
+                    threshold, on the wire — a consumer that needs a different
+                    one is asking for a different verdict and must say so in a
+                    review, not in a local constant.
+
+
+                    `basis_value_exact_tie` — the top two rows are EXACTLY
+                    equal on the quantity the order was made on. A proven
+                    non-separation; no threshold involved.
+
+
+                    `relative_gap_0.10_provisional` — the relative gap
+                    `(first − second) / first` on the basis quantity, against a
+                    0.10 threshold. 🟡 PROVISIONAL: the number is reused from the
+                    repo's one ratified near-tie magnitude (the option-side
+                    `near_tie.threshold`), and the form is RELATIVE because the
+                    basis quantity is normalised by the largest row — which,
+                    after the lever demotion, need not be in the compared pair
+                    at all. A gap between two point estimates is not a
+                    statistical separability test; the ratified statistic will
+                    read ISL's bootstrap (`rank_stability` on this same object)
+                    and will arrive under a DIFFERENT method name.
             rank_stability:
               type: object
               description: >-
--- a/src/assembly/decision-brief.ts
+++ b/src/assembly/decision-brief.ts
@@ -121,6 +121,16 @@ function computeConfigVersion(nSamples: number = STANDARD_N_SAMPLES_DEFAULT): st
 export type BriefAssemblyInput = Pick<RunResponseV3, 'analysis_status' | 'critiques'> & {
   option_comparison?: RunResponseV3['option_comparison'];
   factor_sensitivity?: RunResponseV3['factor_sensitivity'];
+  /**
+   * ⭐ Family-4 S1b (additive, optional): PLoT's ONE canonical driver order and
+   * its attestation. `top_drivers` PROJECTS `ranked_factor_ids` and takes lever
+   * identity from `lever_ids` (the D-U union) instead of re-deriving it from the
+   * response's own `zero_reason` stamp, which under-covers.
+   *
+   * OPTIONAL, and its absence is FAIL-CLOSED: without it `top_drivers` keeps its
+   * pre-S1b stamp-only behaviour. Absence of an attestation is not permission.
+   */
+  driver_order?: RunResponseV3['driver_order'];
   robustness?: RunResponseV3['robustness'];
   m1_coaching?: RunResponseV3['m1_coaching'];
   m1_review?: RunResponseV3['m1_review'];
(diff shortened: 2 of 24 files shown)


Talchain:
## CI verified at the job bytes

**8 of 9 checks pass**, including `test` (9m28s), `gates` on all three OSes, `audit`, `verify`, `guard`, `safety`, `smoke`.

**`validate-structure` — RED, and it is the pre-existing standing red, verified at the job's own output rather than cited from a registry** (CLAUDE.md trap 7b):

- The failing step is **"Validate OpenAPI schema (Redocly)"** — `❌ Validation failed with 37 errors and 28 warnings`. The first step, `tools/validate-openapi-structure.js`, **passes**.
- Pre-proved against a **pristine clone of the base tip `36b3309`** before pushing: baseline **37 errors / 28 warnings**, this branch **37 errors / 28 warnings**, and the **finding set diffs empty**.
- The findings are `error.v1` schema references in the v1 routes (lines 988, 1930, 1935, 2103, 2387 …). This PR's OpenAPI diff touches only `driver_order`'s `separability`/`method` descriptions and the `driver_order` prose block.

`audit` is **green**, so a red audit here would have been real.

Also re-derived rather than cited:
- Branch protection at push time: `HTTP 404 "Branch not protected"` ⇒ `staging` unprotected, the push is the gate.
- `npm run lint`: **86 warnings / 0 errors on the pristine baseline AND on this branch** — unchanged, nothing absorbed.
- Local `npm test`: **6318 passed / 0 failed**; pristine baseline **6256 passed / 0 failed**.

Talchain:
## Review amendments — pushed `816165d25ea2a72601b0c11c60785ea47d5955ae`

All five doc corrections plus the optional nit. **No wire or behaviour change beyond the one-line fail-closed guard.**

| # | Site | Fix |
|---|---|---|
| 1 | `engine-v3.ts` `driver_label` docstring | Rewritten. The two bands now documented as answering different questions and reading different fields: magnitude = this row's `influence_score`; `'biggest'` = `driver_order.ranked_factor_ids[0]`, with the old argmax definition and where the structural argmax still lives (`influence_rank === 1`). |
| 2 | `importance-authority.ts` "NOT DONE HERE, ON PURPOSE" | Rewritten to the post-S1b truth, retitled **CLOSED**, with the amendment citation (§4.3/§8-S1) and each of the three overturned reasons answered — including that the "blast radius zero" census was a trap-12 mirror at moved tips, not inherited. Records that the §4.4 demotion is still in force. |
| 3 | `driver-order.ts` Neil-overrule instruction | `decideProvisionalSeparability` → **`decideSeparability`** (the function that exists). |
| 4 | `driver-label.ts:72` | `indexOfBiggestDriver` → **`indexOfCanonicalTopDriver`**. |
| 5 | PROJECTION REGISTER census | **Derived, not restated** — see below. |
| nit | rival-sign guard | `if (a <= 0 \|\| b < 0) return UNRESOLVED;` — one line, after the exact-tie branch so that path stays byte-unchanged. |

### The census — your finding confirmed at the bytes, then removed as a class

You were right and my sentence understated badly: I named 2 sites, a grep of the stamp-only predicates finds **8+ call sites across 5 modules** (`decision-brief.ts` ×4, `coaching/critiques.ts` ×2, `coaching/headlines.ts` ×2, `routes/v2/run.ts` ×2, `lib/factor-influence.ts` — the last mitigated, the stamp is OR-ed with the structural flag).

It was **wrong at birth, not by decay** — the hand-maintained mirror this repo names as its dominant defect, reproduced inside the register built to prevent it. So rather than fix the sentence I deleted it: the census now lives in **`tests/stamp-only-lever-predicate.census.test.ts`**, which extracts the set from the sources and **fails loud in both directions** — a new surface joining, or one leaving (so a later migration must delete its row and cannot leave the register claiming a debt already paid). Each row carries a checkable reason. The register now points at it and forbids re-stating it in prose.

### New mutants — and one of my own controls was vacuous

| Mutant | Result |
|---|---|
| M14 negative-runner-up guard removed | **1 failed** |
| M15 a new surface joins the stamp-only set | **1 failed** |
| M16 a surface leaves the set | **1 failed** |
| M19 block-comment stripping broken | **1 failed** |
| M20 line-comment stripping broken | **1 failed** |

**Reported against myself:** the census's first discrimination control asserted only that `driver-order.ts` (which mentions the predicates in prose) stays out of the derived set. Mutant **M17 broke the comment stripper entirely and the control stayed green** — no comment in this repo happens to contain the exact form `identifier(`, so it was an absence assertion that could not see a presence (trap 13), guarding the census's own discrimination. Replaced with a control that exercises the stripper directly on synthetic input; it now bites (M19/M20).

A second mutant found **unfalsifiable code in my own helper**: M18 broke the import/re-export stripping and changed nothing, because `import { X } from '…'` can never contain `X(`. Those lines are **deleted**, with the reasoning recorded — defensive code that cannot fail is the same theatre this spec exists to hunt (trap 15).

### Gates at the amended SHA

- `bash scripts/pre-push-validate.sh` — **PASS 6/6**
- `npm test` — **6324 passed / 0 failed** (was 6318; +6 from the census spec and the sign-guard test)
- `npm run typecheck` — clean · `npm run lint` — **86 warnings / 0 errors, unchanged**
- Both mutation worktrees created **outside the repo root** and removed before every gate run; `git worktree list` shows only the main tree.

Noted: the MEDIUM (orchestrator-context `deriveTopDrivers` / analysis-fallback / serialise — the LLM-context sixth surface) is rowed as its own slice, not in this PR.

PR #291, opened 2026-07-30
DO-NOT-MERGE [ROADMAP 2.105] WITHDRAW /v1/counterfactual — typed 501 replaces placeholder arithmetic served behind a real model card

Talchain:
> **DO NOT MERGE — for review.** `staging` is unprotected here (`gh api .../branches/staging/protection` -> HTTP 404), so the push *is* the gate.

## Verdict: HARDEN (typed 501 + caller telemetry), **not** delete

The brief preferred RETIRE. I confirmed nothing live references the route — **and then found this repo already carries a ratified doctrine saying delete is the wrong disposition for exactly this case.** `src/routes/v1/refuse-unavailable.ts`:

> *"Deleting the route would answer 404, which is indistinguishable from a typo and tells an integrator nothing. A typed 501 says: this capability existed, it was withdrawn, here is why … It also keeps the path mounted long enough to MEASURE who was calling it — which is the evidence the deletion decision needs and which a 404 would destroy."*

**Nine routes already ship this way**, with the machinery (`refuseUnavailable`, `REFUSED_ROUTES`, caller telemetry) and an OpenAPI pattern (`/v1/score`, `/v1/sensitivity`) in place. The two earlier *deletions* (`/v1/analysis/sequential`, `/v1/analysis/policy-tree`, `d27fe16`) are the **older** disposition.

It also **strictly dominates deletion on evidence**: my manifest is repo-scoped. It proves UI and CEE hold zero references; it *cannot* prove no external or manual caller exists. Telemetry closes that; a 404 deletes the question.

## The fabrication, verbatim

```ts
// Simulate counterfactual results
const baseline_outcome = intervention.from_value * 100; // Placeholder
const counterfactual_outcome = intervention.to_value * 95; // Placeholder
```

`graph` never read. `seed` and `k_samples` never entered the numbers. **Four layers of real machinery** around two multiplications:

1. a model card asserting *"All else held constant (ceteris paribus)"* and *"No spillover effects"* — causal machinery that never ran;
2. a confidence badge from **hard-coded** `identifiable: true` / `in_linear_range: true`, so identifiability and linearity scored 1.0 and k-coverage 1.0 at the default `k_samples=1000` — **pinned near its ceiling on every path, for inputs it had not looked at**;
3. a determinism stamp that held only because the output was a constant function of the input;
4. `explain_delta` attributions computed **over the placeholder values**.

Genuine work existed (`enforceComputeBudget`, `checkIdentifiability` with a real 400) but **all** of it fed the metadata and **none** touched the estimate. That is the trap: the trustworthy parts were real, so the untrustworthy part read as real.

## Reference manifest — scope stated

Derived with `rg --text` throughout, so NUL-bearing files could not hide (trap 17). Separately verified **zero source files in this repo contain NUL bytes**; note `grep -P` is unavailable on this machine's BSD grep, so `rg`/`grep -a` was the only valid route. Excluded `node_modules/`, `dist/`, `.git/`.

- Handler: 1 file. Registration: exactly one chain (`routes/v1/index.ts:8,93` -> `createServer.ts:1645-46`). No manifest, no glob auto-registration.
- **The route never called `integrations/isl`** — it and `computeCounterfactual` are two *independent* dead ends.
- **No live consumer outside PLoT:** `gh api search/code` for `"v1/counterfactual"` -> **0 hits in DecisionGuideAI, 0 in olumi-assistants-service** at their `staging` tips. The UI's only counterfactual call is ISL-direct (`src/adapters/isl/client.ts` -> `causal/counterfactual/conformal`), corroborating that the real path is UI/CEE -> ISL.

### Two corrections to the brief

- `tests/contract/openapi-runtime-drift.test.ts` is **/v2/run key-level only** and does *not* assert the route set. The gate that actually bites is `tools/openapi-lint-gate.mjs:69`.
- **No UI hook references the dead `/v1/analysis/*` endpoints.** Both are confirmed deleted in `d27fe16` (ancestor of HEAD), zero symbol refs, no spec `paths:` entry. Only prose attestations remain, plus one spec line pointing at *ISL's* endpoint. **Nothing owed.**

## Changed in lockstep

| change | why it was required |
|---|---|
| handler -> unconditional `refuseUnavailable` | new `PLACEHOLDER_ESTIMATE_REASON`, distinct from `FABRICATED_NUMERICS_REASON` (those routes at least varied with the seed; this one didn't vary with the graph) |
| `REFUSED_ROUTES` 9 -> 10 | telemetry test derives from it, so it adapted |
| spec: WITHDRAWN + 401/403/501 only; orphaned components **deleted** | follows the `/v1/score` precedent — no `scoreRequest`/`scoreResponse` survives either |
| `openapi-lint-gate.mjs:69` 429 entry removed | a route refusing unconditionally **cannot produce a 429**; other withdrawn routes are likewise absent |
| Ajv schema + compile + decl + `createValidator` arm deleted | refusal precedes validation |
| demo short-circuit + `getDemoCounterfactualResponse` deleted | a demo payload of a withdrawn capability is still a fabricated counterfactual (the twin hard-coded revenue/LTV behind the same card and badge) |
| `privacy-gate.mjs` probe **retained** | the refusal emits `caller_class` — **new** privacy-relevant log surface. Dropping the probe would remove the route from the sweep exactly when it started logging caller identity |

## Tests — 4 files DELETED, disclosed explicitly

`body.counterfactual.additional-props`, `body.counterfactual.nested-additional-props` (pinned the deleted Ajv schema), `counterfactual.zero-baseline` + its quarantined twin (pinned a `percentage_change` guard over the placeholder values).

`docs/quarantined-tests.md` entries #1 and #6 updated to RESOLVED-BY-WITHDRAWAL so that register (a hand-maintained mirror) doesn't go stale. Worth noting what the quarantine implied and nobody checked: framing it as a *"numeric precision edge case"* treats the surrounding number as real. **The precision of a placeholder is not a numeric-stability question.**

Repointed rather than deleted where a route-level statement still earns its place: `v1-routes` (asserts 501 **and the absence** of `results`/`model_card`/`confidence`/`explain_delta`), `demo-mode`, `demo.shortcircuit`, `input-bounds` (in-range and out-of-range now get the **same** 501), `determinism` (block removed, reasoning recorded: it passed and was meaningless — a constant function of its input is trivially deterministic, so **the fabrication guaranteed the property the test was proud of**).

`analysis-routes.refusal.test.ts` extended to a third finding-category, now covering all **ten** withdrawn routes plus a leak-check that the four credibility fields are absent.

## Mutation check

Faithfully resurrected the route from `staging`: **6 tests RED immediately, and 4 MORE after `npm run build`.** The dist-spawning suites (`v1-routes`, `demo-mode`, `input-bounds`) only bite once the build is redone — which `npm test` does, but an isolated `npx vitest run` does not. **10 biting tests.** Worth knowing for anyone mutation-checking these files.

## Gates

- `npx tsc --noEmit` — exit 0
- `node tools/openapi-lint-gate.mjs` — `GATES: PASS — openapi lint ok + render ok + headers ok`
- **full pre-push gate: 6 checks, 0 failures**
- `npm test` — **6322 passed | 0 failed | 29 skipped**

**Standing-reds delta: EMPTY.** Baseline (separate pristine clone at `3d13e0ac`): 584 files / 6324 passed / **0 failed** / 30 skipped. Every delta accounted for: −3 counted test files (the 4th was already vitest-excluded), −1 skipped (the quarantined case lived inside a deleted file), −2 net tests = −2 body-strictness −3 zero-baseline −1 determinism −1 input-bounds **+5** refusal-suite cases.

## Deliberately NOT done

`ISLService.computeCounterfactual` and its adapter are **not** deleted. Separate dead end, zero production callers, and **already pinned as dead-with-the-wrong-request-shape** by deliberate guards in `tests/isl-request-drift-pairing.contract.test.ts` (it sends `{dag, intervention, target}`; ISL requires `model` + `outcome`). Deleting it would remove the very evidence the doctrine says to preserve, force one-line edits across **78** test files carrying interface stubs, and rewrite two exact-equality inventory assertions — for **zero** reduction in fabrication risk, since nothing calls it. Rowed separately with the full manifest.


Talchain:
## CI follow-up: `validate-structure` — pre-existing standing red, delta now EMPTY

`validate-structure` failed on the first push. Derived rather than assumed, and it is **not mine**:

**The job is path-filtered on `contracts/openapi.yaml`** (`.github/workflows/openapi-sanity.yml`), which is why it did not run on #290 — that PR does not touch the spec. So this PR is simply the first in a while to trigger a job that was already failing.

Measured by running the job's exact command (`npx @redocly/cli lint contracts/openapi.yaml --skip-rule=no-unused-components`) on a **separate pristine clone of `staging` `3d13e0ac`**:

| | errors | warnings | verdict |
|---|---|---|---|
| pristine `staging` | **37** | 28 | FAILS |
| this branch, first push | 37 | **29** | FAILS |
| this branch, now (`7a0b1c8`) | **37** | **28** | FAILS identically |

Full problem-pointer sets are now **byte-identical** — 87 pointers each, `diff` empty.

### The one extra warning was mine, and the fix is the doctrine's own mechanism

It was `operation-2xx-response`: *"Operation must have at least one 2XX response"* — inherent to a route that refuses unconditionally.

`.redocly.lint-ignore.yaml` already held per-pointer exceptions for the three previously-withdrawn routes, plus a written guarantee:

> *"If a FOURTH operation ever loses its success response, Redocly will still fail — which is the behaviour we want to keep."*

`/v1/counterfactual` became that fourth operation and **Redocly did fail, per-pointer and un-absorbed. The mechanism worked exactly as designed.** The entry was added in response to that failure, not pre-emptively, and the prose now says *four* so the guarantee still names the next one.

Why an exception rather than declaring a 2XX, in the file's own words: the rule *"is asking these operations to document a success they can never produce. Declaring a 2XX purely to satisfy the linter would put a false statement into the published contract — precisely the defect class the withdrawal exists to remove."*

### The 37 errors are a tracked, deliberately-unabsorbed gap

They are the `security-defined` errors the ignore file itself documents, with an explicit instruction not to regenerate: *"that would silently absorb the 37 pre-existing security-defined errors too, turning a visible, tracked contract gap into an invisible one."* **I honoured that** — this adds ONE pointer under `operation-2xx-response` and regenerates nothing.

### Bonus: the file independently validates a decision made earlier in this PR

It already records that 429 was deliberately **not** declared for the other withdrawn routes, because *"an undocumented-but-real response is a smaller fault than a documented-but-unverified one."* Demanding a documented 429 from an unconditionally-refusing route would have been exactly the latter — which is why `/v1/counterfactual` was removed from the 429-header list in `tools/openapi-lint-gate.mjs`. That reasoning is now cross-referenced in both places.

Pre-push gate re-run on `7a0b1c8`: **6 checks, 0 failures.**

Talchain:
## Post-review doc round — both nits applied at `eb8bf83`, plus a correction to my own numbers

Each reviewer claim was **re-derived here before acting on it**, not applied on the label.

### Nit 1 — the "post-build" mutant note: my file list was WRONG, the mechanism stands

Re-measured:

| file | `dist/main.js` hits | imports `src/createServer` | sees a src-only mutation without rebuild? |
|---|---|---|---|
| `tests/v1-routes.test.ts` | **1** | 0 | NO — needs the rebuild |
| `tests/demo-mode.test.ts` | **1** | 0 | NO — needs the rebuild |
| `tests/input-bounds.test.ts` | **0** | **1** | **YES — bites immediately** |

**There are TWO real dist-spawners, not three.** `input-bounds` builds the app in-process and does not belong in that list. The warning itself — *an isolated `npx vitest run` after a src-only mutation under-reports, because those suites run the built artefact* — is unaffected and still worth carrying.

Corroboration I already had in-repo and failed to consult: `docs/lanes/LANE13-…md:95` says *"two suites spawn `dist/main.js`"*.

**Where this note lives, and what I did not do:** it is **not in the repo** (`rg` over `src tests tools docs contracts` finds it only in that pre-existing LANE13 doc) — it lived in the commit message, this PR body, and the lane evidence. So the correction lands **here and in the evidence**. I deliberately did **not** force-push to amend the pushed commit message: rewriting a head that has already been reviewed and ratified, for a comment fix, would invalidate the sign-off. **The commit message for `43f8066` still carries the three-file list — this comment supersedes it.**

### Nit 2 — stale count in `.redocly.lint-ignore.yaml` (in-repo, fixed at `eb8bf83`)

`WHEN TO DELETE THIS FILE` still read *"If all three go"* after this PR made it four. I had updated the file's other two count references in the original change and drifted on the third — which is precisely the defect that file is otherwise careful about, so the line now says four **and** tells the next person to update it when adding a fifth pointer.

One count I deliberately left alone: *"The other three withdrawn routes were never in that list"* (the 429 note) is **correct** — it means the three *other than* `/v1/counterfactual`. Re-reading it in context is what stopped me over-applying the fix.

Verified after the edit: 4 pointers under `operation-2xx-response`, prose agrees, and the job's own command still reports **37 errors / 28 warnings — identical to pristine `3d13e0ac`**, so the standing-red delta remains empty. Pre-push gate: **6 checks, 0 failures.**

### My own number, corrected: "87/87 pointer sets" did not name its unit

The reviewer could not reproduce it and gets 72 unique pointers / 65 diagnostics. Both are right about their own unit — I counted **every `at #/…` occurrence including repeats**; they counted **unique pointers**. Neither changes the verdict. Precise restatement: *the redocly problem set is identical between pristine `3d13e0ac` and the head — 37 errors / 28 warnings, with equal pointer multisets (87 occurrences, 72 unique).*

That is the same defect as the ISL count nit on the sibling PR: **a number without its unit invites a reader to think one of you is wrong.** Fixed in both places.

### Also confirmed from the review

`test` passed on `7a0b1c82` and all checks reached terminal state; the only red is the pre-existing `validate-structure`, whose delta the reviewer measured empty **independently on the CI surface I had explicitly declined to claim**. Mutant E bit **8 tests, not 6** — a different mutation shape than mine, and more bite than claimed is not a defect.

PR #295, opened 2026-07-31
fix(isl): retry from remaining budget + bounded Retry-After capture (ROADMAP 2.202 fix ①)

Talchain:
**ROADMAP 2.202, fix ①** — the cheapest repair for the `run_analysis` 500 class.
Diagnosis of record: `PHASE0-EVIDENCE-2026-07-28/diagnosis-run-analysis-500s.md` §4 + §7 FIX 1.
Lane evidence (RED-first, mutations, gate): `PHASE0-EVIDENCE-2026-07-28/fix-plot-retry-budget-2202.md`.

`DO-NOT-MERGE` prefix per the lane brief — review gate, not a merge block on CI.

## The defect

ISL's compute governor caps concurrent analyses **per API key** at `ANALYSIS_WORKERS` (2). PLoT holds
**one key for the whole product**, so that per-caller fairness gate is a *global* 2-slot cap — and one
analysis's ~10 flip probes can occupy both by itself. The 3rd concurrent analysis gets
`429 caller_concurrency_exceeded` back in **133 ms**.

PLoT already classified 429 as retryable (`ISLHttpError.isRetryable`) — and did not retry, because
`/v2/run` chose the base call's attempt count **up front, from worst-case timeouts**:

```
worstCaseMs(2, 60_000) = 121_000  >  ~69_800 remaining  →  1 attempt  →  zero retries
```

So PLoT emitted a typed-failure envelope with **~69.8 s of the request budget unspent**, and CEE
mapped it to the HTTP 500 the tester saw — 4 of 20 chip-click `run_analysis` attempts on the 31 Jul
probe. The retry machinery was correct and **structurally unreachable**.

The flaw was **duration-blindness**: pricing every failure at a full per-attempt timeout treats a
133 ms reject exactly like a 60 s one.

## The fix

Decide **after** the failure, from the budget that **actually remains**: project the next attempt's
real cost (`delay + per-attempt timeout`) and retry iff it still fits. No "is this failure fast?"
constant — every term is derived, and the delay is `max(Retry-After, our backoff)`.

- **`errors.ts`** — `parseRetryAfterMs` (RFC 7231 delta-seconds *and* HTTP-date) and
  `ISLHttpError.retryAfterMs`. ISL sends `Retry-After: 5` on the governor 429 and PLoT **discarded it
  entirely**: response headers were read for `x-request-id` only. Treated as a **minimum** wait — it
  can lengthen the delay, never shorten it below our own backoff.
- **`retry-budget.ts`** (new) — `decideIslRetry`, pure and directly pinnable.
- **`client.ts`** — `ISLRequestOptions.budget`; the retry gate is `decideIslRetry` instead of
  `attempt === maxRetries`; `isl_retry_scheduled` / `isl_retry_declined` telemetry so governor
  contention is observable **without** cross-service log forensics (the diagnosis needed three
  services' Render logs to see this at all).
- **`run.ts`** — the base call passes the configured cap as an *upper bound* plus the live remaining
  budget.

## Preserved deliberately

- A **slow** failure that consumed the budget still gets **no** retry — the property the old up-front
  clamp existed to protect. Pinned wall-clock, with a positive control proving the harness can see a
  retry when one is due.
- An attempt is only ever **started** when its **full** per-attempt timeout still fits, so the base
  call cannot outlive the caller.
- Budget exhaustion still yields the existing typed-failure envelope. No hang, no infinite retry.
- **No widening to non-classified statuses** — a 400 still fails fast after one attempt (pinned).
- **Blast radius is the base call alone**: with no budget supplied the decision degenerates to
  *exactly* the previous behaviour, so flip probes (F3), thresholds (F9) and health are untouched.
  Their existing tests pass unmodified.
- The retry's per-attempt timeout is **not** re-clamped downward to squeeze in an extra attempt —
  that would truncate a slow-but-successful call, the exact harm the base-call clamp's own comment
  says it avoids.

## ⚠ One existing test was RE-POINTED — disclosed, not buried

`tests/plot-remediation.base-call-budget.route.test.ts` asserted
`worstCaseMs(maxRetries, timeoutMs) <= 70_000` **on the up-front attempt count** — it pinned the
duration-blind *mechanism*, not the invariant. Left alone it would have gone RED against its own fix
(`worstCaseMs(3, 60_000) = 183_000`) while appearing to defend it — exactly the mutant the diagnosis
flagged for ISL's fix ②.

The invariant is unchanged and is now enforced **at runtime against the live budget** rather than by
static arithmetic. Its two surviving assertions were GREEN both before and after, so the re-point did
not hollow the file out — and the new wall-clock arms make a stronger claim than the old arithmetic
could.

## Verification

- **RED-first**, pre-fix worktree at base tip `867c92d1`, `src/` untouched: the headline pin fails with
  `expected [ 1785493732893 ] to have a length of 2 but got 1` — **one** request on the wire — and
  `/v2/run` answers `analysis_status: 'failed'`. The live production shape, reproduced.
- **Both positive controls pass pre-fix** (the harness can see a clean run *and* a failure), so the
  success assertions are not vacuous.
- **Three mutants, non-overlapping bites:** remove the budget bound → 9 RED / 3 files (including a
  genuine budget overrun and the client actually beginning to sleep a `Retry-After: 3600`); remove the
  `Retry-After` capture → 3 RED / 1 file; restore the up-front clamp → 4 RED / 2 files.
  All in throwaway worktrees **outside** the clone root (trap 9c), removed before the gate run.
- **Authoritative gate** `bash scripts/pre-push-validate.sh`: **PASSED**, 6/6 checks,
  **6402 passed | 28 skipped, 0 failures**.

## Not done here

**ISL is not touched** — gate 2 queue-vs-reject is fix ② and is rowed separately.
**CEE is not touched** — downstream 429 = `BUSY` rather than `INTERNAL_ERROR` is fix ③ and rides a
CEE lane. This PR removes most of the tester-visible harm; it does not make the residual failure
*honest*, which is fix ③'s job.

### DIFF
--- a/src/integrations/isl/client.ts
+++ b/src/integrations/isl/client.ts
@@ -8,7 +8,7 @@
  * - Error classification
  */
 
-import { ISLHttpError, ISLTimeoutError, ISLNetworkError, isRetryableError, parseRetryAfterMs, type ISLError422 } from './errors.js';
+import { ISLHttpError, ISLTimeoutError, ISLNetworkError, ISLResponseProcessingError, isRetryableError, parseRetryAfterMs, type ISLError422 } from './errors.js';
 import { decideIslRetry, type ISLRetryBudget } from './retry-budget.js';
 import type { ISLHealthResponse } from './types/isl-types.js';
 import { computeOlumiHash } from '../../util/canonical.js';
@@ -170,6 +170,9 @@ export class ISLClient {
       const controller = new AbortController();
       let timeoutId: ReturnType<typeof setTimeout> | undefined;
       let onExternalAbort: (() => void) | undefined;
+      // ROADMAP 2.202 (review item C): true once ISL has returned a 2xx AND its
+      // body has been read — i.e. the compute already happened.
+      let responseCompleted = false;
 
       try {
         timeoutId = setTimeout(
@@ -267,6 +270,13 @@ export class ISLClient {
         // Lane PLoT-R3 (2.13): read the raw text first so responseDigest covers
         // the EXACT bytes received, then parse (same JSON, same errors as .json()).
         const responseText = await response.text();
+        // ROADMAP 2.202 (review item C): past this point ISL has COMPUTED the
+        // analysis and handed us the bytes. Anything that fails from here is a
+        // PLoT-side processing fault, and re-issuing the analysis cannot fix
+        // it — see the catch below. `response.text()` stays OUTSIDE the flag
+        // because a socket reset mid-body is a genuine transport failure and
+        // IS worth retrying.
+        responseCompleted = true;
         const responseData = JSON.parse(responseText) as T;
         const islEchoedRequestId = response.headers.get('x-request-id') ?? null;
         const responseHash = computeOlumiHash(responseData);
@@ -305,6 +315,18 @@ export class ISLClient {
         } else if (rawError instanceof ISLHttpError) {
           // Already typed
           wrappedError = rawError;
+        } else if (responseCompleted) {
+          // ROADMAP 2.202, review item C — NARROWED.
+          //
+          // ISL already returned 2xx and we already have its bytes, so the
+          // analysis HAS been computed; a failure here (JSON.parse, hashing,
+          // bookkeeping) is ours. The generic `else` below wraps it as an
+          // ISLNetworkError, which isRetryableError reports as RETRYABLE — and
+          // 2.202 is precisely what makes those retries reachable. Left alone,
+          // a post-compute parse failure would re-issue the WHOLE analysis and
+          // multiply load on the very governor whose contention this change
+          // exists to survive.
+          wrappedError = new ISLResponseProcessingError(endpoint, rawError);
         } else {
           // Unknown error - wrap as network error
           wrappedError = new ISLNetworkError(endpoint, rawError);
--- a/src/integrations/isl/client.ts
+++ b/src/integrations/isl/client.ts
@@ -8,12 +8,15 @@
  * - Error classification
  */
 
-import { ISLHttpError, ISLTimeoutError, ISLNetworkError, isRetryableError, type ISLError422 } from './errors.js';
+import { ISLHttpError, ISLTimeoutError, ISLNetworkError, isRetryableError, parseRetryAfterMs, type ISLError422 } from './errors.js';
+import { decideIslRetry, type ISLRetryBudget } from './retry-budget.js';
 import type { ISLHealthResponse } from './types/isl-types.js';
 import { computeOlumiHash } from '../../util/canonical.js';
 import { recordIslSuccess, recordIslFailure, shouldAllowIslCall } from '../isl-circuit-breaker.js';
 import { recordDownstreamCall, sanitizePayloadForDebug, computePayloadDigest } from '../../util/downstream-tracker.js';
-import { ISL_TIMEOUT_MS, ISL_HEALTH_CHECK_TIMEOUT_MS, resolveIslMaxRetries, islRetryBackoffMs } from '../../config/timeouts.js';
+// ROADMAP 2.202: the backoff series now lives behind decideIslRetry (which
+// still derives it from islRetryBackoffMs — the single source is unchanged).
+import { ISL_TIMEOUT_MS, ISL_HEALTH_CHECK_TIMEOUT_MS, resolveIslMaxRetries } from '../../config/timeouts.js';
 
 /**
  * ISL client configuration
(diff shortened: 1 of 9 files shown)


Talchain:
## Amendment round — adversarial review addressed

Head `3330db3e` → **`b79c48298edd46ff09045b783636a31dc8e816ab`**. CI **11/11 pass** on the new head. Gate `bash scripts/pre-push-validate.sh`: **PASSED, 6/6, 6409 passed | 28 skipped | 0 failed** (was 6402 — +7 net new arms).

### A1 (HIGH) — fixed
`decideIslRetry` computed `delayMs = max(retryAfterMs, backoffMs)` **before** the `!budget` branch and returned it there, where by definition there is no budget to bound it — in a sleep that observes no `AbortSignal`. The `!budget` path now ignores `retryAfterMs` and sleeps its own backoff (capped at `ISL_RETRY_BACKOFF_CAP_MS`). Only the budgeted path may honour `Retry-After`, because only there is the delay checked against anything.

The reviewer's live-exposure finding was correct and is the reason this was a HIGH, not a nit: `/v1/run` calls `validateCausal` + `analyseRobustness` with `maxRetries = 3` and **no budget**, against the very endpoint whose governor 429 carries `Retry-After: 5` — a ~+7s regression per call under exactly the contention this PR fixes.

**RED-first vs `3330db3e`, verbatim:**
```
× ⭐ IGNORES Retry-After entirely — an unbounded hint must not reach an unbounded path
  → expected 3600000 to be 1000
× the no-budget delay is bounded by our backoff CAP for every attempt
  → expected 86400000 to be 1000
× ⭐ NO BUDGET — Retry-After is ignored; the client sleeps its own backoff (A1)
  → expected 3007 to be less than 2000
× ⭐ NO BUDGET — an absurd Retry-After cannot stall the call (A1)
  → Test timed out in 30000ms.
```
That timeout is the hour actually being slept on the real client — empirical, not inferred.
**Mutant** (restore the unbounded delay): 2 RED, `expected 3600000 to be 1000`.

### A2 — pinned
The block titled *"no budget supplied = previous behaviour exactly"* had **no arm passing `retryAfterMs`**, so its title asserted something no test could see. Added a unit arm (absurd **and** the realistic 5s governor hint), a cap-bound arm across attempts 1–5, and two client-level wall-clock arms on the real socket.

### A3 — two false claims in the evidence file corrected, not silently rewritten
Both were true only of the **budgeted** path while written as properties of the whole change:
1. *"an absurd `Retry-After` self-limits… no arbitrary cap needed"* — self-limiting is a property of **the budget**, not the parser.
2. *"no budget = EXACTLY the previous behaviour"* — false pre-A1; the pre-2.202 client slept 1s where the first cut slept 5s. **This one concealed a live regression**, so it was not merely imprecise.

Recorded with the lesson in the platform's own terms: I proved a property on the budgeted path and generalised it to a path that has no budget by definition — *a capture proves what it was pointed at.*

### B1 — taken
Literals `70_000`/`60_000` replaced with `ISL_TIMEOUT_MS` / `resolveRequestBudgetMs()`, plus an explicit headroom pin naming the three-way cross-repo coupling. Live headroom is only **~8.9s**.
**Mutant** (`ISL_TIMEOUT_MS` 60s→64s, one line): **both** the headroom arm *and* the functional `RETRIES the observed 133ms 429` arm go RED — so the guard catches a genuine disablement, not just arithmetic:
```
→ 2.202 is DISABLED at these constants: a 5000ms Retry-After plus a 64000ms attempt plus the
  1000ms margin does not fit 70000ms of budget. Re-derive ISL_TIMEOUT_MS / REQUEST_BUDGET_MS together.
```

### C — TAKEN (not rowed), and it was measured
Against `3330db3e`, a fake ISL returning `200` with an unparseable body: **`expected 3 to be 1`** — the analysis was **recomputed three times** after ISL had already computed it, multiplying load on the very governor this PR exists to survive. Narrowed via a non-retryable `ISLResponseProcessingError`, scoped by a `responseCompleted` flag. **`response.text()` deliberately stays outside the flag** — a socket reset mid-body is a genuine transport failure and *is* worth retrying. Positive control: same server, same config, retryable 429 → 2 hits.

Disclosed side-effect: `/v1/run`'s `maxRetries = 3` callers also stop retrying post-2xx parse faults. It is a *narrowing*, in the direction we want.

### D — disclosed
`base_isl_call_budget_clamped`'s trigger narrowed, so it **goes silent at the default budget** (it previously fired on every base call via its second disjunct). Accepted because the information it carried — the attempt ladder — was the thing that was wrong, and `isl_retry_scheduled` / `isl_retry_declined` carry strictly more at the moment of decision. Flagged for anyone grepping for that row in future Render forensics.

### E — softened
*"impossible by construction"* was an overstatement: `BASE_CALL_MIN_TIMEOUT_MS` floors the per-attempt timeout, so at a small `REQUEST_BUDGET_MS` the **first** attempt can exceed what remains. Restated to what the code gives — no **retry** starts unless its full per-attempt timeout fits.

Full transcripts: `PHASE0-EVIDENCE-2026-07-28/fix-plot-retry-budget-2202.md` (amendment section).

PR #317, opened 2026-08-07
feat(2.855): forward goal_constraints[].value_frame to ISL at every rebuild (2.798)

Talchain:
**Row 2.855** — the transport half of row 2.798:
`ISL declares ✅ → CEE stamps (CEE #862) → **PLoT forwards** → a constraint delivers`.
**Do not merge without independent review** — trust surface. Pairs with **Talchain/olumi-assistants-service#862**.

## Why

ISL's fail-closed reader (2.796) is deployed: a constraint with no `value_frame` is refused (`CONSTRAINT_FRAME_UNSPECIFIED`) and the whole `constraint_analysis` block is omitted. PLoT had **zero** occurrences of the field across all 328 `.ts` files under `src/`, so nothing CEE stamps could reach the engine.

## ⚠ The field dies at THREE rebuilds, not one

Ingress is **permissive** — the route's `goal_constraints` items carry no `additionalProperties: false`, so AJV's `removeAdditional` never fires, and `V2_RUN_ALLOWED_KEYS` is top-level only. The strip is **downstream**, at three by-presence literals that keep only the keys they name:

1. `normalisation/constraint-filter.ts` — **always** on path; its output *replaces* the whole list
2. `lib/intervention-normaliser.ts` — conditional
3. `integrations/isl/translator-v3.ts` — the ISL egress

All three carry it now. **Absent in ⇒ absent out**: PLoT never mints, defaults or infers a frame. Normalising against a node's range is a change of **scale, not frame**, so the attestation rides through unchanged.

## ⚠ Declared on `engine-v3.ts`, not only on the wire type

`tools/gen-structural-keys.mjs` reads `engine-v3.ts` but **not** `translator-v3.ts`. Declaring only on `ISLGoalConstraint` would pass the structural-keys drift gate **green** while the key silently became unregistered decision content (digested by `pii-redact`, tokenised by `decision-tokens`). Declaring it here made the gate **fail loud**; registry regenerated in this commit.

## Evidence — through the real endpoint, because a green build proves nothing here

ISL's model is `extra: "ignore"`: a misspelt key returns a clean 200 with no error anywhere. Five arms captured against `POST /api/v1/robustness/analyze/v2` at ISL commit `c695feb7`, frozen in `tests/fixtures/isl-constraint-value-frame-20260807/`:

| arm | request | outcome |
|---|---|---|
| **A** control | no frame | 200, 4 results, **zero** `constraint_analysis`, `CONSTRAINT_FRAME_UNSPECIFIED / frame_not_stamped` |
| **B** valid | `'delta'` | 200 and a **real computation** — every option: `prob_satisfied 0.856`, `joint_probability 0.856`, `failure_margin_median`, `near_miss_fraction`, `binding` |
| **C** invalid | `'normalised'` | **HTTP 422**, `analysis_status "blocked"`, typed reason naming `goal_constraints -> 0 -> value_frame` |
| **D** misspelled | `value_frmae` | 200 and **nothing** — byte-identical in outcome to A |
| **E** level | `'level'` | 200, no result, but `CONSTRAINT_NOT_CONVERTIBLE / missing_target_baseline` |

**D is the load-bearing arm.** Without it, B is a 200 being read as a computation. The test pins D's misspelling *in-test*, so the discrimination cannot silently stop discriminating (trap 13b).

Fixtures are pinned **by content at a named commit** (trap 12b) — evidence about `c695feb7`, and they say so.

## ⚠ `'level'` still cannot deliver — and this PR does not claim it can

Arm E is the honest limit. ISL recovers a level as `baseline + (option_sample − status_quo_sample)`, so it needs `observed_state.baseline` **on the constraint's target node**. PLoT forwards a baseline but **never mints one**, and ROADMAP 2.281 covers the **goal** node only — a constraint target is usually a factor/outcome node. **New gap to row, not a defect here.**

The discrimination that matters: **the refusal reason moved.** A frame-blind reading of "still no result" would call this a failure; in fact the frame was consumed and the chain advanced to the next precondition.

## ⚠ Scope of the witness

These are a **local-build** witness at `c695feb7`, **not** a deployed-staging witness — capturing against `isl-staging` needs an API key this lane did not have and would not go hunting for. **The deployed arm is still owed** before the chain may be called live.

## Mutation check

Reverting the source to pristine `staging` REDs **6 of 8** forward arms. The two that stay green are the *"absent stays absent"* arms — correctly, since absence **is** pristine behaviour; they guard a future *default*, not the field's presence. Applied-check control on pristine `src/` read **exactly 0**. Worktree outside the repo root, isolation proven **by writing** a sentinel (source hash unchanged, distinct inodes), removed before the gate run.

## Gate at this tip

`npm run lint` · `npm run validate:contracts` · `npm run build` · `RATE_LIMIT_ENABLED=0 npm test` — **all exit 0**; 629 files, 7,056 tests, 0 failed. No existing key-set pin or egress fixture moved (by-presence means an absent key never appears).


Talchain:
### `audit` is RED — pre-existing, empty delta from this PR (read at the job output, not inherited from a registry)

Per trap 7b I re-read the failing job rather than restating a known-red's cause from the registry. It reports:

```
fast-uri  3.0.0 - 3.1.4 || 4.0.0 - 4.1.1
Severity: high
fast-uri vulnerable to host confusion via backslash authority introducer
  — GHSA-7p8r-x3mc-p8w7
node_modules/@fastify/ajv-compiler/node_modules/fast-uri
node_modules/ajv/node_modules/fast-uri
node_modules/fast-json-stringify/node_modules/fast-uri
node_modules/fast-uri
5 high severity vulnerabilities
```

A transitive production advisory reached via `fastify` / `ajv` — the one already rowed as **ROADMAP 2.385**.

**Proof the delta is empty, by construction rather than by assertion:** this PR's diff against `origin/staging` touches **zero** dependency-surface files — no `package.json`, no `package-lock.json`, no shrinkwrap. A dependency advisory cannot be introduced by a change that alters no dependencies. (`audit` does not run on `staging` pushes, only on PRs, so a direct base-run comparison is not available — hence the structural argument rather than a colour comparison.)

Every other check passes: `test` · `verify` · `guard` · `gates` (macos/ubuntu/windows) · `safety` · `smoke`. The three contexts required on `main` — `test`, `verify`, `guard` — are all green.

Talchain:
ADVERSARIAL REVIEW (paired with CEE #862) — VERDICT: MERGE-WITH-NOTE. Merge order: THIS PR FIRST; #862 must NOT follow until the P0 below is resolved.

Fresh blobless clone, HEAD asserted at 33e07d9fe1d096d439b740549d0cfaeaa5778812 before AND after all reads; tree clean at exit.

WHY THIS IS SAFE TO MERGE ALONE
Inert without a producer. PLoT never mints, defaults or infers a frame, and no CEE branch stamps one today, so absent-in => absent-out and nothing changes for any user. Ingress verified not to reject an undeclared key (run.ts:1320-1323 has no item-level additionalProperties:false), so ordering either way cannot 4xx.

CONFIRMED
- Tarball byte-identity verified across ALL THREE repos: git blob 4fc13289886dbc1b73a28aa80b9d54518a3dc4dd, 353,278 bytes, sha256 761c7ec6... (CEE, PLoT, DGAI@staging).
- Arm D (the load-bearing one) is pinned IN-TEST — `expect(d.value_frmae).toBe("delta")` — inside a test that pins its own precondition, so the discrimination cannot silently stop discriminating.
- Local-build scope is disclosed prominently in both the test header and the commit message. Nothing describes the chain as live or deployed.
- Gates re-run independently: npm run lint 0 · validate:contracts 0 (8 files) · build 0 · RATE_LIMIT_ENABLED=0 npm test 0 (629 files, 7,056 passed).
- `audit` red proven PRE-EXISTING BY CONSTRUCTION, not by label: `git diff stagingref pr317 -- package.json package-lock.json` is EMPTY, so the dependency surface is unchanged. The red is the 5 fast-uri high advisories of ROADMAP 2.385.

🚨 P0 FOUND — DO NOT LAND #862 UNTIL RESOLVED (the harm appears only when the SECOND PR lands)
The reduction branch is the ONLY delta producer and it mints a NEGATIVE value. PLoT clamps negatives to zero while faithfully forwarding the delta attestation.
intervention-normaliser.ts:1521 — the negative sign is ITSELF the normalisation trigger: `if (constraint.value < 0 || constraint.value > 1) { return true; }`
Measured against this clone own module:
  constraintsNeedNormalisation([-0.15]) = true
  normaliseValue(-0.15, [0,1]) = {normalised:0, clamped:true}   (same for [0,100] and [0,500000])
  WHAT REACHES ISL: {operator:"<=", value:0, original_value:-0.15, value_frame:"delta"}
run.ts:6528 -> :6845 -> the translator sends c.value (the clamped 0), never original_value. ISL delta path returns GoalThresholdPlan(delta_threshold=threshold) with NO warning.
NET: the user asks "P(cost falls by >=15%)" and the product computes "P(cost falls at all)".
Aggravating: no PLoT test anywhere covers a negative constraint value, and intervention-normaliser.ts:1459 has ZERO branch coverage (68 evaluations, true branch taken 0 times) — nothing would red.
Smallest correct fixes: (a) PLoT fails CLOSED — drop a delta-framed constraint that clamped, keeping ISL all-or-nothing refusal honest; or (b) #862 lands with the reduction branch stamping NOTHING, lighting the six level branches while the delta path stays honestly dark.

ALSO REFUTED — the new comment at intervention-normaliser.ts claims normalising is "a change of SCALE, not of FRAME: a level stays a level and a delta stays a delta". The transform is AFFINE WITH AN OFFSET (value - min), and a non-zero min is reachable (inferred_spread, lines 276/615). A delta does not survive an offset. The comment describes the behaviour the code should have, not the behaviour it has.

NOTES TO CLOSE (recorded, not blocking this merge)
R1. The ISL egress drift gate is STRUCTURALLY BLIND to value_frame: the pinned ISL model (686fcb7f) does not declare it, no egress producer emits it, and it is not in knownUndeclared. Probe with positive control: control -> UNEXPLAINED=[]; same body + value_frame -> ["goal_constraints[].value_frame"]. This reproduces, one field over, the coverage hole this repo own isl-egress-producers.ts:299-310 documents.
R2. Rebuild site 2 (lib/intervention-normaliser.ts) has ZERO coverage; the mutation kit (4 hop-1 + 4 hop-3 arms) was structurally incapable of biting it. The header claim "all THREE carry it now" is unsupported by the kit.
R3. There is a FOURTH by-presence rebuild at canonicalise.ts:370-377 that DROPS value_frame. Not on the ISL path, but two runs differing only by frame now share a response_hash while producing different results.
R4. Stale claims this PR falsifies and leaves standing: run.ts:5567-5572 and :5671 assert ISLGoalConstraint has no frame field (citing the very range this PR extended); vendor/README.md:27 states a grep result that is now false.
R5. contract.test.ts:98 `describe("... the deployed handler CONSUMED the field")` contradicts the file own scope note 53 lines above. Wording fix.
R6. PLoT eslint does NOT have no-dupe-keys: eslint.config.js:1 imports @eslint/js and never spreads it, so only 5 rules resolve. Pre-existing, worth its own row — CEE has it as an error, PLoT does not.

PR #318, opened 2026-08-07
[P0 2.878] refuse a 'delta' constraint whose value normalisation would alter — PLoT clamped -0.15 to 0 and forwarded the attestation

Talchain:
**ROADMAP 2.878 (P0)** + **2.879 (4)**. Pairs with **Talchain/olumi-assistants-service#862**, which was declined to merge over exactly this.
**Do not merge without independent review** — trust surface. ⚠ **Read the lint note at the bottom before merging.**

## ⚠⚠ THIS PR IS A PRECONDITION, NOT A FIX FOR THE LIVE CORRUPTION — read this first

**Today, on staging, PLoT still clamps `-0.15` → `0` and forwards it.** This PR does not stop that, and anyone reading it as closing the live corruption will be wrong.

The guard keys on `value_frame === 'delta'`, and **that attestation does not exist yet**. Re-derived at the bytes:

| fact | evidence |
|---|---|
| CEE `staging` `b5204544` has **zero** `value_frame` occurrences | repo-wide grep at that tip |
| `extractor.ts:451` mints `value: -value` **unattested** | source at that tip |
| `translator-v3.ts:950` forwards by presence and **never invents a frame** | source at this tip |

So the reduction constraint still normalises to `0` and is still forwarded, exactly as before — because the attestation the guard needs is precisely what is absent.

**What this PR does is close the seam BEFORE #862 opens it.** #862 stamps `'delta'` on every reduction branch, at which point *every* decision containing a "reduce X by N" phrase would hit this path. Landing this first means the fabrication never becomes reachable. **Blast radius today: zero for CEE-originated traffic** (nothing is stamped) — though PLoT's `/v2/run` ingress does not validate `value_frame`, so any direct client stamping it reaches this code now.

---

## The defect, reproduced by execution against this module

CEE's `extractReductionConstraints` mints `{ operator: '<=', value: -N }` for *"reduce X by N"* and attests it `'delta'`. PLoT then:

```
normaliseGoalConstraints([{ value: -0.15, value_frame: 'delta', operator: '<=' }], [costNode])
  → constraint out: { operator: '<=', value: 0, original_value: -0.15, value_frame: 'delta' }
  → diagnostic:     { normalised_value: 0, clamped: true, range: {0, 200, inferred_value} }
```

The translator sends `value: c.value` — the clamped **`0`** — and the `'delta'` attestation with it. **ISL is asked `<= 0` in the delta frame: "what is the probability cost changes by at most nothing", i.e. the probability cost falls AT ALL.** It answers with a confident `prob_satisfied` and no warning, because a frame-faithful `0` is a perfectly well-formed delta. `original_value` is a PLoT-internal field the translator never sends, so nothing downstream can recover the question.

**Corrected premise (one).** The row says `-0.15` → `0` *"for every range"*. Measured: for every range whose **floor is 0** — `default`, `unit_percent`, `goal_threshold_cap`, `explicit_cap`, and `inferred_value` on any positive-domain node, i.e. the overwhelming majority. Against a **sign-preserving** range like `[-1, 0]` (D-9) it becomes **`+0.85`**, *not clamped and not flagged* — a second, independent corruption the row did not name, and one a `clamped`-based guard would miss entirely. Both are closed here.

> ⚠ **An earlier version of this body said `+0.425` here. That was wrong** — `0.425` is the `[-1, 1]` case; `[-1, 0]` gives `0.85`, which is what the spec has always asserted. The error propagated from this body into a register row and a dispatch brief before review caught it. Corrected in all four places.

## The clamp's original purpose, as derived — and why it is left alone

`normaliseFiniteValue` computes `(v − min) / (max − min)` and clamps to `[0,1]`. For a **LEVEL** — an absolute position on a factor's own bounded scale, which is what every other producer mints — that is correct, and the clamp is a **domain guard**: ISL's sample space is `[0,1]` by construction, so a level outside the derived range must be pinned to the nearest endpoint. `clamped: true` is the honest signal that the derived range was too narrow. **Downstream depends on exactly that** — `constraintThresholdClampByConstraintId` reads the clamp *direction* so the margin egress never labels an understated `failure_margin_median` 'exact'.

It is category-wrong for a **DELTA** in two ways: the translation term `− min` is meaningless on a *difference*, and the `[0,1]` clamp deletes the sign, which is the entire content of a reduction.

**So the two paths are separated, and neither is weakened.** The clamp is untouched; three PIN tests hold it in place (level below floor → 0 + forwarded, level above ceiling → 1 + forwarded, unattested → unchanged).

## The chosen outcome: PLoT fails closed, per-constraint

Of the two recorded options, **"PLoT fails closed on clamp"**, widened to the invariant itself.

**Why not "the reduction branch stamps nothing":** that is a *CEE* change, it permanently retires the only constraint class that delivers today, and **it fixes nothing here** — an unattested negative value is still clamped to `0` and still forwarded, and any other producer that later stamps `'delta'` (the `add_constraint` handler and client ingress are both live, unconstrained paths) re-opens it. That option closes one producer; this closes the seam.

**Why the predicate is `normalised !== value` and not `clamped`, the sign, or a range-source list:**

| candidate guard | what it misses | mutant |
|---|---|---|
| `clamped` | the un-clamped translation corruption (`-0.15` → `+0.425`), and the rescaled positive delta | **M2 → RED** |
| `value < 0` | the entire positive half of the delta domain (`+0.05` → `0.00025`) | **M3 → RED** |
| a list of "safe" range sources | a hand-maintained mirror a new ladder rung silently escapes | — |
| **`normalised !== value` ALONE** | **the affine map's FIXED POINTS** — `[0,0.5]` with delta `1` clamps `2`→`1`, so `normalised === value` while the quantity is HALVED | **R4 → RED** |

**⚠ The shipped predicate is the UNION `(normalised !== value \|\| clamped)`, and an earlier version of this PR was wrong to claim the outcome test "cannot be escaped".** Review refuted it by execution: the two guards have **disjoint blind spots** and neither dominates. Both arms are load-bearing — removing either reopens a class (R4 and R5 below).

No false positives from floating point: the only exact-identity path is `min = 0 ∧ width = 1`, where `(v − 0) / 1 === v` exactly in IEEE-754, so a faithfully-forwarded delta is never refused.

> ⚠ **A second record correction.** An earlier version called `+0.05` → `0.00025` "a 200× understatement". **That presupposes the raw-unit reading, which this same PR declares unestablished** — for any `min = 0` range, `(v − min)/w ≡ v/w`, i.e. the shipped number *is* the arithmetically correct **normalised-space** delta. The substitution is **unverifiable in either direction**, and that is exactly why failing closed is right. Arguing it with a magnitude claim that assumes the semantics is the same over-read this fix exists to prevent. (The **clamp**, by contrast, is a corruption under *both* readings: no reading makes `<= 0` mean `<= −0.15`.)

**Why refuse rather than rescale.** Correctly rescaling a delta means dividing by the range *width* with no offset — but whether ISL's `delta` frame expects a raw-unit or normalised-space difference **is not established at the bytes**. The only witness we hold (arm B of `tests/fixtures/isl-constraint-value-frame-20260807`) sent `0.05` already on the normalised scale. Minting arithmetic on an unverified semantics is the fabrication class this chain exists to remove. When ISL's delta scale contract is established, this guard is where the conversion goes — and it will still be the thing that refuses what it cannot convert faithfully.

**Why drop rather than throw.** A throw would take out every option result over one constraint. The refusal is per-constraint and **reported by name**: a `CONSTRAINT_REFUSED_FRAME_FIDELITY` preflight warning, a `'removed'` repair record, an entry in `_meta.filtered_constraints`, and a `plot.constraint_refused` log event. *Dropping the `value_frame` instead was considered and rejected* — ISL fail-closes an unstamped constraint by omitting the **entire** `constraint_analysis` block, so one bad constraint would silently delete every other constraint's verdict, and it would still send the corrupted number.

## The input space is bounded by the producer, not by me

Trap 16-inverse — *a fixture you wrote yourself is not evidence about the wire.* The fixture reproduces what CEE actually mints, read at `olumi-assistants-service@2d29951b`, `src/cee/compound-goal/extractor.ts:468-485`. **One detail that changes which code is reachable:** its `unit` is **`"fraction"`, not `"%"`** — `normaliseConstraintUnits` (ROADMAP 1.52) relabels a `%` unit whose `0 < |value| < 1`. `isPercentUnit('fraction')` is **false**, so the `unit_percent` `[0,100]` rung is *not on this path*; a fixture using `"%"` would have tested a rung the producer cannot reach. Constraint shape cross-checked against the real captures in `l60-artefacts/scenario-{people,pricing}.json` — which are also the evidence that every live constraint today is positive and level-shaped, and thus why this sat unobserved.

## Evidence at this head (`c5305b5`)

**RED-first.** Module suite at pristine: **16 red / 8 green** (6 red on the defect; 6 more red only because `result.refused` did not exist yet — labelled **`PIN:`** rather than shipped as `DEFECT:` theatre; the 8 green are the `constraintsNeedNormalisation` branch tests, correctly green because that trigger was never wrong and simply had **zero** coverage). Route suite: reverting only the F1 fix REDs **3 of 9**.

**Mutation — 13 mutants across three rounds, every one as predicted, no survivors.** Throwaway worktree outside the repo root each round, **isolation proven by WRITING** a sentinel (distinct inodes, source sha256 unchanged), applied-check scoped to `src/` with the **control reading exactly ZERO**, collected count constant within each round.

| # | mutant | expected | result |
|---|---|---|---|
| M1 | guard deleted | RED | **RED** ×8 |
| M2 | predicate → `clamped` only | RED | **RED** ×2 |
| M3 | predicate → `value < 0` | RED | **RED** ×1 |
| M4 | frame flipped to `'level'` | RED | **RED** ×10 |
| M5 | `continue` removed | RED | **RED** ×6 |
| **M6** | **narrowed to an UNRELATED id** | **GREEN** | **GREEN** ✅ |
| **M7** | **narrowed to exclude the NAMED id only** | RED | **RED** ×5 |
| R1 | F1 fix removed | RED | **RED** ×3 |
| R2 | F1 filter inverted | RED | **RED** ×3 |
| **R3** | **filter keyed on an id never refused** | **GREEN** | **GREEN** ✅ |
| R4 | `\|\| clamped` arm removed | RED | **RED** ×1 |
| R5 | outcome arm removed | RED | **RED** ×2 |
| R7 | critique suppressed | RED | **RED** ×1 |
| M-F3c | new log site, different event, record by reference | RED | **RED** ×1 |
| M-F3d | same, nested two levels deep | RED | **RED** ×1 |
| **M-F3e** | **new log site with only SAFE fields** | **GREEN** | **GREEN** ✅ |

**R4 and R5 red DIFFERENT tests** — that is the evidence neither predicate arm is redundant. **M6/M7 and R3** are the identity-discrimination pairs. **M-F3c/M-F3e** prove the log guard bans quantities rather than new log sites.

⚠ **Two guards of mine failed under their own mutants and were rebuilt — disclosed because it is the more useful half.** `R6` (whole record logged by reference) **survived** a name-scanning guard; the replacement was site-scoped and then **M-F3c survived that**. The guard now binds to the **emission** — `createServer` wires `formatters.log = redactLogRecord`, called for every record from every site, so wrapping it and capturing its return value observes what actually ships, regardless of site or event name. A site-scoped guard is a mirror of where the code is today.

## Gates at this tip

| gate | result |
|---|---|
| `npm run lint` | ✅ 0 |
| `npm run typecheck` | ✅ 0 |
| `npm run validate:contracts` | ✅ 8 files / 3 scenarios, 0 failed |
| `npm run build` | ✅ 0, no stale `.js` |
| `RATE_LIMIT_ENABLED=0 npm test` | ✅ **631 files / 7,081 passed**, 0 failed, 28 skipped |

`#317`'s baseline was 629 files / 7,056 tests. This branch adds **1 file / 20 tests**; the full-suite figure above was measured when the eslint gate test was still on the branch (+2/+25) — see the split note below. **CI is the authority for this tip.**

**`dependency-audit` — proven pre-existing, not assumed.** PLoT's `staging` does not run the `audit` workflow, so it is not usable as a control; **I ran the exact CI command (`npm audit --omit=dev --audit-level=high`) myself against staging tip `1b36c34a` as the baseline** and against this branch. Output is **byte-identical**: `5 high severity vulnerabilities`, all `fast-uri` `GHSA-7p8r-x3mc-p8w7` (row 2.385). This PR touches **no** dependency file — `git diff staging..HEAD -- package.json package-lock.json` is empty. **Zero delta.**

## ⚠ Split — this PR is the P0 ONLY

This branch originally also carried a fix to `eslint.config.js` (ROADMAP 2.879 (4): `@eslint/js` was imported and never spread, so only **5 rules** resolved and `no-dupe-keys` was **absent** — the rule that would have caught the duplicate object key which silently overrode `'delta'` with `'level'` in this same chain).

Making the linter able to see brought **7 pre-existing findings** into view and turned `npm run lint` red, which **gated the test step so CI never ran the suite**. That was the wrong trade for a P0: it would have shipped a live-fabrication fix on a *local* test number.

**The eslint change now lives in its own PR (linked below), with those 7 findings listed and untouched.** They are not swept — two may be deliberate (`no-control-regex` on what looks like the intentional NUL sentinel, and `no-async-promise-executor`) and are being rowed rather than cleaned under cover of a config change.

**This branch is `c13a49c` alone and is lint-green.**

## What this does NOT claim

- **Not a deployed witness.** Unit + module level only; no staging run drove a refusal.
- **Delta constraints do not start delivering.** Today CEE stamps nothing, so nothing is refused in production either — blast radius is **zero at the current deploy state**, which is exactly why this should land *before* #862, not after.
- **`'level'` is still blocked** one hop on, at ISL's `missing_target_baseline` (ROADMAP 2.877). Untouched here.

### DIFF
--- a/src/lib/intervention-normaliser.ts
+++ b/src/lib/intervention-normaliser.ts
@@ -1209,6 +1209,43 @@ export interface ConstraintNormalisationResult {
     /** True when the node's CEE-stamped goal_threshold was preferred (P0-C1) */
     used_node_goal_threshold?: boolean;
   }>;
+  /**
+   * ROADMAP 2.878 — constraints REFUSED rather than forwarded, because
+   * normalisation would have changed the number a `delta` attestation stands
+   * behind. Empty on every path that does not hit that case. See
+   * `DELTA_FRAME_VALUE_ALTERED` for the full reasoning.
+   */
+  refused: RefusedConstraintRecord[];
+}
+
+/**
+ * ROADMAP 2.878 — the single reason PLoT refuses to forward a constraint it
+ * successfully normalised. A union of one, deliberately: a new refusal reason
+ * must be added HERE and every consumer's exhaustiveness check moves with it,
+ * rather than a bare string that can accrete meanings silently.
+ */
+export type ConstraintRefusalReason = 'delta_frame_value_altered_by_normalisation';
+
+/** ROADMAP 2.878 — see {@link ConstraintRefusalReason}. */
+export const DELTA_FRAME_VALUE_ALTERED: ConstraintRefusalReason =
+  'delta_frame_value_altered_by_normalisation';
+
+/**
+ * ROADMAP 2.878 — a constraint that PLoT declined to send to ISL, with the
+ * quantities that made the refusal necessary. Shaped so it can be widened into
+ * `FilteredConstraintRecord` (`{constraint_id, node_id, reason}`) for
+ * `_meta.filtered_constraints` without a second hand-copy.
+ */
+export interface RefusedConstraintRecord {
+  constraint_id: string;
+  node_id: string;
+  reason: ConstraintRefusalReason;
+  /** The quantity the user stated, exactly as it arrived. */
+  stated_value: number;
+  /** The number normalisation would have sent in its place. */
+  would_have_sent: number;
+  /** The range whose scale produced that substitution. */
+  range: NormalisationRange;
 }
 
 /**
@@ -1259,6 +1296,8 @@ export function normaliseGoalConstraints(
   const repairs: RepairRecord[] = [];
   const diagnostics: ConstraintNormalisationResult['diagnostics'] = [];
   const normalisedConstraints: NormalisedGoalConstraint[] = [];
+  // ROADMAP 2.878 — constraints refused rather than forwarded with an altered value.
+  const refused: RefusedConstraintRecord[] = [];
 
   // Build node lookup map
   const nodeMap = new Map<string, EngineNodeV3>();
@@ -1414,6 +1453,94 @@ export function normaliseGoalConstraints(
       usedNodeGoalThreshold = true;
     }
 
+    // ROADMAP 2.878 — FRAME FIDELITY. A constraint attested `delta` reaches ISL
+    // as the quantity the user stated, or it does not reach ISL at all.
+    //
+    // WHY THIS IS NOT "the clamp is wrong". `normaliseFiniteValue` computes
+    // `(v − min) / (max − min)` and clamps to [0,1]. For a LEVEL — an absolute
+    // position on a factor's own bounded scale, which is what every other
+    // producer mints — that is correct and the clamp is a DOMAIN GUARD: ISL's
+    // sample space is [0,1] by construction, so a level outside the derived
+    // range must be pinned to the nearest endpoint, and `clamped: true` is the
+    // honest signal that the derived range was too narrow to contain the stated
+    // level. Downstream depends on exactly that (see
+    // `constraintThresholdClampByConstraintId` in routes/v2/run.ts, which reads
+    // the clamp DIRECTION so the margin egress never labels an understated
+    // failure margin 'exact'). The clamp is right, and it is left alone.
+    //
+    // It is category-wrong for a DELTA, in two independent ways, both measured:
+    //   1. the translation term. `− min` is meaningless on a DIFFERENCE. A
+    //      delta of −0.15 against a sign-preserving range [−1,1] normalises to
+    //      **+0.425** — not clamped, not flagged, and not the user's quantity.
+    //   2. the clamp itself. Against every range whose floor is 0 — which is
+    //      `default`, `unit_percent`, `goal_threshold_cap`, `explicit_cap`, and
+    //      `inferred_value` on any positive-domain node, i.e. the overwhelming
+    //      majority — a negative delta normalises to **0** for EVERY width.
+    //      `−0.15` → `{normalised: 0, clamped: true}`.
+    //
+    // Case 2 is the P0 this guard exists for. `extractReductionConstraints`
+    // mints `{operator: '<=', value: −N}` for "reduce X by N" and attests it
+    // `delta`; the clamp turns `<= −0.15` into `<= 0`, so ISL is asked
+    // *"P(cost changes by at most 0)"* — **the probability cost falls AT ALL** —
+    // and answers it with a confident `prob_satisfied` and no warning, because a
+    // frame-faithful `0` is a perfectly well-formed delta. Nothing downstream
+    // can recover the question: `original_value` is carried on the PLoT-internal
+    // struct and is NOT a field the ISL translator sends.
+    //
+    // THE PREDICATE IS THE INVARIANT ITSELF — "the number that would reach ISL
+    // is not the number the user stated" — and deliberately NOT `clamped`, nor
+    // the value's SIGN, nor the range's source. `clamped` is one CAUSE of the
+    // harm and would miss case 1 entirely; the sign is what the defective
+    // `constraintsNeedNormalisation` gate already keys on and would miss a
+    // rescaled positive delta; a source list would be a hand-maintained mirror
+    // that a new ladder rung silently escapes. An outcome test cannot be escaped
+    // by any of those. Note it also passes the identity cases for free: the
+    // forward-raw branch and a [0,1] range both leave `normalised === value`, so
+    // a faithfully-forwarded delta is never refused.
+    //
+    // WHY REFUSE RATHER THAN RESCALE. Rescaling a delta correctly means
+    // dividing by the range WIDTH with no offset — but whether ISL's `delta`
+    // frame expects a raw-unit difference or a normalised-space difference is
+    // NOT established at the bytes. The only witness we hold (arm B of
+    // tests/fixtures/isl-constraint-value-frame-20260807) sent `0.05` already on
+    // the normalised scale. Minting an arithmetic on an unverified semantics is
+    // precisely the fabrication class this chain exists to remove, so PLoT
+    // refuses and says so. When ISL's delta scale contract is established, this
+    // guard is where the conversion goes — and it will still be the place that
+    // refuses anything it cannot convert faithfully.
+    //
+    // WHY DROP RATHER THAN THROW. A throw would take out the whole analysis —
+    // every option result — over one constraint. The refusal is per-constraint:
+    // the run proceeds, every other constraint delivers, and this one is
+    // reported by name in `_meta.filtered_constraints` with a `removed` repair
+    // record. Dropping the `value_frame` instead was considered and rejected:
+    // ISL fail-closes an unstamped constraint by omitting the ENTIRE
+    // `constraint_analysis` block, so one bad constraint would silently delete
+    // every other constraint's verdict — and it would still send the corrupted
+    // number, merely unattested.
+    if (value_frame === 'delta' && normalised !== value) {
+      refused.push({
+        constraint_id,
+        node_id,
+        reason: DELTA_FRAME_VALUE_ALTERED,
+        stated_value: value,
+        would_have_sent: normalised,
+        range,
+      });
+      repairs.push({
+        field: `constraint.value.${constraint_id}`,
+        action: 'removed',
+        from_value: value,
+        to_value: 'refused',
+        reason:
+          `refused: a 'delta'-framed constraint value cannot be normalised without ` +
+          `changing the quantity it attests (range=[${range.min},${range.max}] ` +
+          `source=${range.source} would have sent ${normalised} in place of ${value}). ` +
+          `Forwarding it would ask ISL a different question than the user asked.`,
+      });
+      continue;
+    }
+
     // UNIT COMPATIBILITY (the goal-fit unit collision). Decided HERE, at
     // ladder-decision time, for the same reason range_unified is: this is the
     // only place that holds BOTH the constraint's own declared unit and the
@@ -1505,6 +1632,7 @@ export function normaliseGoalConstraints(
     constraints: normalisedConstraints,
     repairs,
     diagnostics,
+    refused,
   };
 }
 
--- a/src/lib/intervention-normaliser.ts
+++ b/src/lib/intervention-normaliser.ts
@@ -1478,6 +1478,20 @@ export function normaliseGoalConstraints(
     //      majority — a negative delta normalises to **0** for EVERY width.
     //      `−0.15` → `{normalised: 0, clamped: true}`.
     //
+    // ⚠ WHAT IS *NOT* CLAIMED, and an earlier draft of this comment got wrong.
+    // For a min-0 range the map reduces to `v / width`, which IS the
+    // arithmetically correct NORMALISED-SPACE delta — so `+0.05` on `[0,200]`
+    // → `0.00025` is only a corruption under the RAW-UNIT reading, and this
+    // very file states below that which reading ISL takes is NOT established.
+    // The substitution is therefore **unverifiable in either direction**, not a
+    // measured 200× error. That is precisely why refusing is right: under
+    // genuine uncertainty a gap beats a number nobody can check. Stating it as
+    // a settled corruption would be the same over-read this guard exists to
+    // prevent (an honest unknown hardening into a claim the next reader
+    // inherits without its scope).
+    // Case 2's CLAMP, by contrast, is a corruption under BOTH readings: no
+    // reading of a delta frame makes `<= 0` mean `<= −0.15`.
+    //
     // Case 2 is the P0 this guard exists for. `extractReductionConstraints`
     // mints `{operator: '<=', value: −N}` for "reduce X by N" and attests it
     // `delta`; the clamp turns `<= −0.15` into `<= 0`, so ISL is asked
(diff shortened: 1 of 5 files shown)


Talchain:
## ⚠ CI status — read before triaging

**3 red, and all three are accounted for:**

| check | cause | verdict |
|---|---|---|
| `test` | **Lint (ESLint)** step only — the 7 disclosed pre-existing findings | expected; ruling owed |
| `verify` | **Lint (ESLint)** step only — same 7 | expected; ruling owed |
| `audit` | 5 high `fast-uri` (`GHSA-7p8r-x3mc-p8w7`), ROADMAP 2.385 | **pre-existing, proven** |

Read at the bytes in both job logs — the failing step in `test` and `verify` is `Lint (ESLint)`, nothing else:

```
185:39  error  Unexpected control character(s) in regular expression: \x00, \x1f  no-control-regex
 46:13  error  Unnecessary escape character: \$   no-useless-escape
 51:11  error  Unnecessary escape character: \$   no-useless-escape
 52:17  error  Unnecessary escape character: \$   no-useless-escape
  8:22  error  Promise executor functions should not be async  no-async-promise-executor
 17:45  error  Unnecessary escape character: \/   no-useless-escape
 17:56  error  Unnecessary escape character: \"   no-useless-escape
✖ 7 problems (7 errors, 0 warnings)
```

**`audit` is proven pre-existing, not assumed.** PLoT's `staging` does not run that workflow, so it is not usable as a control. I ran the exact CI command (`npm audit --omit=dev --audit-level=high`) against **staging tip `1b36c34a` as the baseline** and against this tip: **byte-identical output**. `git diff staging..HEAD -- package.json package-lock.json` is **empty**.

### ⚠ Scope limitation, stated rather than glossed

**Lint gates the test step, so CI never ran the suite on this branch.** The `631 files / 7,081 tests passed` in the PR body is my **local** run of `RATE_LIMIT_ENABLED=0 npm test` at this exact tip — not a CI witness. Dropping the eslint commit (below) would let CI run and produce one.

### The lever

The two changes are **separate commits on purpose**:

- **`c13a49c`** — the P0 fix. **Lint-GREEN on its own**, verified by checking it out alone in an isolated worktree.
- **`85fe0c5`** — the eslint config fix. This is what turns lint red, by making the linter able to see code it has been blind to since the config was written.

**Drop or defer `85fe0c5` and everything goes green.** I did not sweep the 7 or suppress them, per the brief — the sweep is a ruling, and one of them (`errors.ts`'s deliberate NUL sentinel, trap 17) is a judgement rather than a cleanup.

Talchain:
## ✅ CI WITNESS — the suite actually ran, and it is green

Split applied: this branch is **`c13a49c` alone**. The eslint config change moved to **#319**.

**`test` job at `c13a49c`** — [run 31189842044](https://github.com/Talchain/plot-lite-service/actions/runs/31189842044/job/92903572141). Every step succeeded, and critically **`Run tests` executed** rather than being gated out by lint:

```
success  Check for stale .js files in src/
success  npm ci
success  Lint (ESLint)          ← green, so it no longer gates the suite
success  Validate contracts
success  npm run build
success  Run tests              ← RAN
```

```
Test Files  630 passed | 4 skipped (634)
     Tests  7076 passed | 28 skipped (7115)
  Duration  167.27s
```

**The arithmetic confirms the split is exactly what was removed.** My local pre-split run measured `631 files / 7,081 tests`; dropping the eslint gate test (1 file / 5 tests) gives `630 / 7,076` — the CI numbers, exactly. Nothing else moved.

### Check state

| check | result |
|---|---|
| `test` | ✅ **suite ran, 7,076 passed, 0 failed** |
| `verify` | ✅ |
| `gates` (ubuntu / macos / windows) | ✅ ✅ ✅ |
| `guard` · `safety` · `smoke` · `update_release_draft` | ✅ |
| `audit` | ⚠️ **pre-existing, proven** |

**`audit` is the only red, and it is the standing `fast-uri` one (row 2.385).** Read in the CI job log at this SHA:

```
fast-uri  3.0.0 - 3.1.4 || 4.0.0 - 4.1.1
fast-uri vulnerable to host confusion via backslash authority introducer — GHSA-7p8r-x3mc-p8w7
  node_modules/@fastify/ajv-compiler/node_modules/fast-uri
  node_modules/ajv/node_modules/fast-uri
  node_modules/fast-json-stringify/node_modules/fast-uri
```

Proven pre-existing rather than assumed: PLoT's `staging` does not run that workflow, so it is not usable as a control — I ran the exact CI command (`npm audit --omit=dev --audit-level=high`) against **staging tip `1b36c34a` as the baseline** and against this branch, and the output is **byte-identical**. `git diff staging..HEAD -- package.json package-lock.json` is **empty**. **Zero delta from this PR.**

**Ready for review. Not merged.**

Talchain:
## Adversarial review addressed — `bb606c6`, CI green with the suite executed

Every finding accepted. No pushback on any of them; the two record errors were mine and are corrected in the body, the code comments, and here.

### F1 — BLOCK, fixed

**Remedy taken: remove the refused ids from `activeGoalConstraints` at the point of refusal**, mirroring what the temporal filter does by replacing the list before the active set is derived.

Chosen over "pass `constraintsForISL` to the response builder" because `activeGoalConstraints` has **eight** downstream consumers, and fixing it at the source repairs all of them at once — **including F9**, which the alternative would have left live. A refused constraint no longer reaches `buildConstraintScaleProvenance`, so it can no longer disclose as a forwarded-raw `source: 'default'` constraint that never travelled.

⚠ **The reviewer's framing is the part worth keeping in the record:** this fix reproduced the exact harm its own comment cited as the reason for rejecting the alternative design. That comment now carries the counter-example and points at the route-level test, so the next reader cannot re-derive the same hole.

### F2 — predicate widened

`(normalised !== value || clamped)`. The dominance claim was refuted by execution and I have withdrawn it: **identity of the output does not imply identity of the transform** — `[0,0.5]` with delta `1` clamps `2`→`1`, leaving `normalised === value` while halving the quantity. The two guards have disjoint blind spots. Both arms are pinned by mutants that red *different* tests (R4 vs R5).

### F3 — leak closed, and the guard rebuilt twice

Dropped `stated_value` / `would_have_sent`; flattened `range` → `range_source`. Quantities remain only in the `removed` repair record, which is response content rather than an info log.

Re-anchored the two redaction pins from a fixed **1,300-byte window** to **brace-matched extraction**, and added a **derived enumerator** over every `req.log` site in `run.ts`.

⭐ **Then a mutant proved my own new guard did not bite.** R6 restored the leak (`refused: result.refused`) and the enumerator **survived** — it scans for quantity field *names*, and logging a whole typed record by reference never spells them. A guard agreeing with itself. Added a second guard asserting the value is a **key projection onto an allow-list**, derived from the source; R6 now reds.

⚠ **Scope stated rather than implied:** that is still a source-level check on one site. The general property needs a runtime capture of the real `/v2/run` logger, and `createServer` exposes no stream injection point — pino writes to fd 1 via sonic-boom, where a naive capture reads zero bytes and every absence assertion passes vacuously. **Not attempted rather than faked.**

### F4 — the untested half now has tests and mutants

`tests/constraint-delta-frame-refusal.route.test.ts`, 8 tests through the real route. **The ISL mock derives its results from the constraints it actually received** rather than returning a fixed list — a hardcoded mock returns N results regardless of what PLoT forwarded, which is precisely the blindness that let the count mismatch ship.

That also fixes the criticism of the old wire test: absence is now asserted on the **real ISL payload**, with the siblings asserted **present** in the same loop so it cannot pass vacuously.

**RED-first**: reverting only the F1 fix REDs 3 of the 8.

### Mutation — 7/7 as predicted, control exactly ZERO

Throwaway worktree outside the repo root, isolation proven by writing (distinct inodes `561024324`/`561034387`, sentinel landed, source sha256 `77b3ac55…` unchanged). Collected count constant at **49** on every run.

| # | mutant | result |
|---|---|---|
| R1 | F1 fix removed (the shipped BLOCK) | **RED** ×3 |
| R2 | F1 filter inverted | **RED** ×3 |
| **R3** | **filter keyed on an id that is never refused** | **GREEN** ✅ |
| R4 | `\|\| clamped` arm removed | **RED** ×1 (the fixed-point witness *only*) |
| R5 | outcome arm removed, `clamped` alone | **RED** ×2 (translation + positive-delta) |
| R6 | F3 leak restored | **RED** ×1 *(survived until the guard was rebuilt)* |
| R7 | refusal disclosed but critique suppressed | **RED** ×1 |

R4 and R5 redding **different** tests is the evidence that neither predicate arm is redundant.

### F5 — record corrected

- `[-1,0]` gives **`0.85`**, not `0.425` (that is the `[-1,1]` case). Corrected in the body and comments.
- The **"200× understatement"** is withdrawn: it presupposes the raw-unit reading this same file declares unestablished. For any `min = 0` range the map *is* the correct normalised-space delta. Restated as **unverifiable in either direction** — which is why failing closed is right. The **clamp** remains a corruption under *both* readings.
- Stated plainly that **the live corruption persists until #862 lands**: `extractReductionConstraints` mints `value: -N` **unattested** today, and the guard keys on the attestation.

### F6 / F7 — rowed, not fixed here, and **not** resolved incidentally by F1

I checked both against the F1 fix specifically, since you asked:

- **F6** (the refusal drops essentially the whole delta class on in-graph nodes — identity needs `min = 0 ∧ width = 1`): **not** resolved. F1 changes the *blast radius* of a refusal, never *which* constraints are refused. Now recorded in-source as **"delta-framed constraints are non-functional until ISL's delta scale contract is established"**, so it cannot be inherited as an edge case. Re-surface trigger: that contract.
- **F7** (the same constraint delivers or is refused depending on an unrelated sibling, via the global `constraintsNeedNormalisation` gate): **not** resolved — the gate is upstream of everything F1 touches.

### Gates at `bb606c6`

| gate | local | **CI** |
|---|---|---|
| `lint` | ✅ 0 | ✅ |
| `typecheck` / `build` / `validate:contracts` | ✅ 0 | ✅ |
| full suite | ✅ 631 files / **7,089** passed | ✅ **631 files / 7,089 passed** |

**CI witness — the suite executed** ([run 31193874430](https://github.com/Talchain/plot-lite-service/actions/runs/31193874430/job/92917197125)): `Lint (ESLint)` ✅ → `Validate contracts` ✅ → `npm run build` ✅ → **`Run tests` ✅**. `test`, `verify`, `gates` ×3, `guard`, `safety`, `smoke` all green. CI and local agree exactly.

**`audit` is the only red** — 5 high `fast-uri` `GHSA-7p8r-x3mc-p8w7` (row 2.385), proven byte-identical against staging tip `1b36c34a`; no dependency file is touched.

**Not merged. Ready for re-review.**

Talchain:
## Round-2 fixes applied — `c5305b5`, CI green with the suite executed

### 1. The third leak shape — guard now binds to the EMISSION

M-F3c is the sharpest instrument finding in this PR, and it is the same lesson as R6 one level out. Both of my earlier guards were **mirrors of where the code is today**: guard A scanned for quantity *names* (blind to a record logged by reference); guard B asserted a key projection at *one event name* (blind to a new site under a different name).

The property is not "this site is safe" — it is **"no log line this service emits ever carries a decision quantity"**. So the guard now watches the emission: `createServer` wires `formatters: { log: redactLogRecord }`, which pino calls for the root logger **and every child**, i.e. for every record from every site. Wrapping it and capturing its **return** value observes exactly what ships to the sink, after redaction, regardless of site or event name. It deep-walks keys and values, and pins its own precondition (capture non-empty **and** containing the site under test) so it cannot go vacuous.

**Discriminating triple:**

| mutant | expected | result |
|---|---|---|
| **M-F3c** new site, different event, whole record by reference | RED | **RED** ✅ |
| **M-F3d** same, nested two levels deep under an innocuous key | RED | **RED** ✅ |
| **M-F3e** new site emitting only SAFE fields | **GREEN** | **GREEN** ✅ |

M-F3e is the load-bearing arm: it proves the guard bans **quantities**, not **new log sites**.

### 2. The "200×" withdrawal now reaches the spec

`tests/constraint-delta-frame-fidelity.test.ts:268` still asserted it as fact. Fixed, with the reasoning in place rather than a deletion — and the test's own point restated as being about the **predicate's domain** (a sign-keyed guard would not fire there at all), not about a magnitude. Swept the tree: the only remaining occurrences of "200x" and "0.425" are inside the withdrawal text itself, and the `0.425` in `intervention-normaliser.ts:1474` is correct — it names `[−1,1]`, while the spec's `0.85` names `[−1,0]`. **The error only ever existed in this PR body.**

> The reviewer's line is the one worth keeping: *the correction reached the two documents a reviewer reads and missed the one a maintainer will.*

### 3. PRECONDITION banner added

The body now opens with it, re-derived at the bytes: CEE `staging` `b5204544` has **zero** `value_frame` occurrences, `extractor.ts:451` still mints `value: -value` unattested, `translator-v3.ts:950` never invents a frame, and the guard requires the attestation. **Today PLoT still clamps `-0.15` → `0` and forwards it.** This PR closes the seam *before* #862 opens it; it does not close the live corruption.

The body is also brought current at this head — the full 16-mutant table now appears, so the R4/R5 forward-references resolve.

### Also done

Corrected the plan-count exactness comment at `run.ts:~5980`. Recorded honestly: the divergence is a **pricing** one, not a correctness one (the request still carries exactly the PUs for the constraints it forwards); measured at **zero** so far; **reachability UNKNOWN**, needing a refused delta that is the only constraint on its target node — with the settling experiment named rather than a defect asserted.

### Gates at `c5305b5`

| gate | local | **CI** |
|---|---|---|
| lint / typecheck / build / contracts | ✅ | ✅ |
| full suite | 631 files / **7,090** passed | ✅ **631 files / 7,090 passed** |

**CI witness — suite executed** ([run 31198534751](https://github.com/Talchain/plot-lite-service/actions/runs/31198534751/job/92932607396)): `Lint` ✅ → `Validate contracts` ✅ → `build` ✅ → **`Run tests` ✅**. `test`, `verify`, `gates` ×3, `guard`, `safety`, `smoke` all green.

**The SSE timing flake did not appear** — clean first run, no retries.

**`audit` remains the only red**: 5 high `fast-uri` `GHSA-7p8r-x3mc-p8w7` (row 2.385), byte-identical to staging tip `1b36c34a`; no dependency file touched.

**Not merged.**

PR #323, opened 2026-08-08
fix(2.957): a %-unit threshold must mean the same thing alone and batched

Talchain:
> **⚠ This PR was reversed in direction after adversarial review.** The first revision read `unit:'%'` as always meaning percentage points and rescaled the sub-1 cell. That was **refuted at the producer's bytes** — it would have turned `0.04` (four percent) into `0.0004` on CEE's primary draft path, i.e. it would have *introduced* a 100× error on the one path that was still correct. The defect it aimed at is real; the fix was pointed the wrong way. Rebuilt below. The post-mortem is at the end, kept deliberately.

## The defect (ROADMAP 2.957)

`normaliseGoalConstraints`' `unit_percent` rung resolved `[0,100]` **unconditionally**, while the route only invokes the normaliser when `constraintsNeedNormalisation` fires (`value < 0 || value > 1`) or some node carries a non-identity intervention scale. So the same constraint got **two different meanings depending on an unrelated constraint's value**:

| Payload | Wire value sent to ISL |
|---|---|
| `{unit:'%', value:0.04}` **alone** | `0.04` ✅ correct |
| the **same row** + any out-of-range batch-mate | **`0.0004`** ❌ 100× low |

One constraint, two meanings, decided by a *different* constraint. That is indefensible whichever convention wins — which is why the row stands even though its original direction did not.

## Which arm is wrong is settled by the PRODUCER, not by PLoT

CEE's LLM extractor emits **`"4%"` as `value: 0.04, unit: "%"`** — a *fraction* under a `'%'` label. It says so on the very function written to stop a consumer misreading it (`olumi-assistants-service`, `src/cee/compound-goal/extractor.ts:925-934`):

> *"The LLM extractor converts "4%" to value: 0.04, unit: "%". If a consumer interprets unit: "%" as "value is percentage points", **0.04% ≠ 4%**. … Rule: if unit === "%" and 0 < value < 1 → already fractional → unit = "fraction"."*

It **relabels without dividing**, and it runs **only on the regex branch** (`stages/repair/compound-goals.ts` — the LLM-emitted branch below it does not call it), so a fractional value under a raw `'%'` label reaches PLoT on the primary draft path.

The other half of the rule is CEE's too, pinned in its own suite (`tests/unit/cee.constraint-unit-normalisation.test.ts`):

> *"preserves percentage-unit constraints where value >= 1 (already in pp form)"* — `// value: 4, unit: "%" means "4 percentage points"`

## The fix

The rung now reproduces **CEE's boundary exactly** — transcribed, not invented:

| Constraint | Reading | Range | Change |
|---|---|---|---|
| `\|v\| < 1` | already fractional | identity `[0,1]` — forward as-is | **fixed** (was `v/100`) |
| `\|v\| >= 1` | percentage points | `[0,100]` — divide by 100 | **unchanged** |

Plus one narrow conjunct closing the last batch-dependent cell: `value === 1` satisfies neither gate bound, so a lone `{unit:'%', value:1}` — *one* percentage point — was forwarded raw as `1.0` (a *hundred* percent) yet resolved to `0.01` once batched. Every other percentage-point magnitude already opens the gate, so the blast radius is exactly that cell.

**One shared predicate** (`isPercentPointValue`) serves the route disjunct, the forward-raw exemption and the rung's complement, so the three cannot drift. A first cut had them disagree at `value === 1` — the route forced an invocation the forward-raw rung then no-opped — caught by the batch-invariance sweep.

**The F1 pin (`'%' 0.5 → 0.5`) is UNTOUCHED.** Under this direction it is correct as written, which is itself corroboration: `constraint-intervention-range-unify.test.ts` does not appear in this diff at all.

## The property: batch-invariance

**The value PLoT sends for a constraint is a function of that constraint alone.** Pinned as a sweep across both sides of CEE's boundary *and* both sides of the legacy gate's boundary (`0, 0.004, 0.04, 0.9, 1, 4, 40, 100`), each run twice — with and without a gate-opening batch-mate — asserting the two agree.

## ⚠ Behaviours that MUST NOT change — declared, with fixtures

| # | Must-not-change behaviour | Fixture |
|---|---|---|
| 1 | `40%` → `0.4` (percentage points still rescale) | `PIN: 40 with unit '%' is FORTY PERCENT` — *passes at pristine too*, by design |
| 2 | A non-percent sub-1 row is still forwarded raw, no repair/diagnostic | `CONTROL: a NON-percent sub-1 row …` |
| 3 | `goal_threshold_cap` (3) still outranks `unit_percent` (4) | `CONTROL: rung precedence … (3)` |
| 4 | A measured non-identity intervention scale (1) still wins | `CONTROL: rung precedence … MEASURED …` |
| 5 | `unit_percent` (4) still outranks an identity scale (5) | `CONTROL: rung precedence … IDENTITY …` |
| 6 | `>100%` still clamps to 1 with `clamped: true` | `CONTROL: a >100 %-unit row still clamps` |
| 7 | A negative percentage-point delta still resolves on `[0,100]` | `CONTROL: a negative percentage-point delta` |
| 8 | `isPercentUnit` unchanged (rejects `count`, `months`, `fraction`) | `CONTROL: isPercentUnit unchanged` |
| 9 | **The F1 forward-raw parity pin, in full** | `constraint-intervention-range-unify.test.ts` — **not in this diff** |

Percent-token handling is **derived from `PERCENT_UNIT_TOKENS`** and swept over every token, fractional and percentage-point.

## Evidence

**RED-first at pristine `747a3aa4`:** 6 failed / 10 passed. The failures are the fractional and batch-invariance pins; the `40 → 0.4` control passes at pristine *by design* — it guards an arm this PR does not move.

**Route-level wire witness.** The invocation decision lives in `run.ts`, so a module suite is structurally blind to it. The route suite asserts the value ISL *actually receives*, and **pins its own gate-closed precondition** (a non-percent row observed being forwarded raw — possible only when both legacy disjuncts are false), so no arm can pass for an unrelated reason.

**Mutation kit** — worktree outside the repo root, isolation proven by **sentinel write** (source SHA unchanged) + inode comparison; applied-check scoped to `src/`; control exactly zero; collected count matches pristine every run.

| Mutant | Verdict | Killed by |
|---|---|---|
| CONTROL (pristine) | GREEN | baseline, 43 collected |
| **M1** sub-1 `'%'` never fractional (**the refuted direction**) | **RED** 6 | the `0.04` pins, module **and** wire |
| **M2** `'%'` always fractional (never rescale pp) | **RED** 12 | `40 → 0.4`, `1 → 0.01`, and pre-existing `'%'` coverage |
| **M3** drop the route disjunct | **RED** 2 | *only* the wire tests — module suite green |
| **M4** drop the forward-raw pp exemption | RED 3 | the `value === 1` cell + batch-invariance sweep |
| **M5** `>= 1` → `> 1` (off-by-one at the producer boundary) | RED 4 | the boundary + invariance pins |
| **M6** hardcode the `producerDeclaredRange` mirror | GREEN → **RED** 1 | see below |

**M1/M2 are the discriminating pair**, and they now straddle the *actual* question: M1 is the direction this PR was reversed from, M2 is its opposite. Both bite, on different assertions. **M3** proves the wire witness is not redundant with the module suite.

**M6 survived the first kit, and was measured rather than assumed equivalent.** It is *not* equivalent: with a measured `[0,100]` intervention scale and a fractional `{unit:'%', value:0.04}`, the producer scale is `[0,1]`, so the threshold sits on a different scale than its samples — a real divergence that must deny the decision-grade trust bit. Hardcoded, the two compare equal and the divergence is silently certified as unified. Closed with a test **and its discriminating twin** (a percentage-point value against the same scale *is* unified), so the new assertion cannot pass by always reporting divergence.

**Full named gate** (`npm test`) at `f9aa7657`, via the repo's own pre-push validation: **`Tests 7142 passed | 28 skipped (7181)` — 0 failures.** `tsc --noEmit` clean.

**No changes to `package.json`, `vendor/`, or the lockfile.** Rebased onto current `staging` (`747a3aa4`), so this is tested against the **schemas 0.39.0** pin that actually ships.

## Interaction with ROADMAP 2.958 — explicit ruling

**Unchanged for `|v| >= 1`, and this PR now moves 2.958 in the *right* direction for `|v| < 1`.** The percentage-point arm is byte-identical to before. The fractional arm stops PLoT applying a percentage-points reading to a value CEE emits as a fraction — which is precisely the cross-service convention skew 2.958 records. The CEE half remains out of scope for this lane.

**Residual, recorded not fixed** (both confirmed by the reviewer):
- The `constraint-compiler.ts` path pairs `observed_state.value` with the node's top-level `unit`; for a true-90% node it now yields `0.009` in **both** gate states. The "already broken when the gate is open" claim is verified — but this converts a correct-by-accident cell into a consistently wrong one. Not fixed here; it belongs with 2.958's compiler half.
- `parseValue`'s `'%/month'` suffix escapes both `isPercentUnit` and the new predicate. **Pre-existing**, untouched.
- A ratio that legitimately exceeds 100% (NRR `110%` → `value: 1.10, unit:"%"` on a *node*) sits on the ambiguous side of CEE's `>= 1` constraint rule. This PR **does not resolve it** — it reproduces CEE's stated constraint rule exactly, so PLoT never invents a third convention.

## Post-mortem — why the first revision was wrong

I derived the convention from PLoT's own pinned contract (*"Threshold in the user's units; PLoT normalises downstream"*) plus the existence of a percent→fraction rewrite, and concluded that a surviving `'%'` label meant raw percent. **That contract line does not settle which scale `'%'` denotes**, and the rewrite does not run on the LLM path. The evidence that settles it lives in the CEE repo — **which my own PR body named as a location I could not reach**.

Two lessons, both already in the estate's doctrine and both re-earned here:

1. **A disclosed unreached location is where the refutation lives.** I disclosed it correctly and then reasoned as though the gap were immaterial. The disclosure is not the mitigation.
2. **A full mutant kill-rate proves sensitivity, never correctness** (trap 13c). The first revision passed 6/6 against an oracle derived from the consumer's contract rather than the producer's behaviour — a perfect score on the wrong exam. Every expectation in this revision is transcribed from CEE's source and its own test suite, with file and line.

### DIFF
--- a/src/lib/intervention-normaliser.ts
+++ b/src/lib/intervention-normaliser.ts
@@ -1099,6 +1099,118 @@ export function isPercentUnit(unit: string | undefined): boolean {
 }
 
 /**
+ * ROADMAP 2.957 — THE PERCENT SCALE IS DECIDED BY THE PRODUCER'S CONVENTION,
+ * AND THAT CONVENTION IS MAGNITUDE-DEPENDENT. THIS IS NOT PLoT'S CHOICE.
+ *
+ * ⚠ READ THE PRODUCER BEFORE CHANGING THIS. A `'%'` constraint does NOT always
+ * carry percentage points. CEE's LLM extractor emits **`"4%"` as
+ * `value: 0.04, unit: "%"`** — a FRACTION under a `'%'` label — and CEE's own
+ * `normaliseConstraintUnits` exists to relabel exactly that case, with a
+ * comment naming the opposite reading as the defect it prevents
+ * (`olumi-assistants-service/src/cee/compound-goal/extractor.ts:925-934`):
+ *
+ *     "The LLM extractor converts "4%" to value: 0.04, unit: "%".
+ *      If a consumer interprets unit: "%" as "value is percentage points",
+ *      0.04% ≠ 4%.  This normaliser detects the fractional case and relabels
+ *      the unit to "fraction" so the convention is unambiguous.
+ *      Rule: if unit === "%" and 0 < value < 1 → already fractional."
+ *
+ * It relabels WITHOUT dividing — the value was already fractional. And the
+ * relabel is NOT guaranteed to have happened: `normaliseConstraintUnits` is
+ * called only on the REGEX-extracted branch
+ * (`stages/repair/compound-goals.ts` — the LLM-emitted branch below it does
+ * not call it), so a fractional value under a raw `'%'` label reaches PLoT on
+ * the primary draft path.
+ *
+ * THE OTHER HALF OF THE PRODUCER'S RULE, equally load-bearing: a `'%'`
+ * constraint whose value is `>= 1` IS percentage points. CEE pins this itself —
+ * `tests/unit/cee.constraint-unit-normalisation.test.ts`:
+ *   "preserves percentage-unit constraints where value >= 1 (already in pp
+ *    form)" · "// value: 4, unit: "%" means "4 percentage points""
+ * and `value === 0` is preserved as `'%'` (0 is 0 on either reading).
+ *
+ * So the boundary below is CEE's boundary, transcribed — not a heuristic:
+ *   |v| <  1  ⇒ already fractional ⇒ IDENTITY [0,1]; forward the value as-is.
+ *   |v| >= 1  ⇒ percentage points  ⇒ [0,100]; divide by 100 (UNCHANGED
+ *               behaviour — this is what the rung has always done).
+ *
+ * ⚠ WHAT THIS FIXES. The rung previously used [0,100] unconditionally, so
+ * `{unit:'%', value:0.04}` — the producer's own canonical example, meaning 4% —
+ * became **0.0004** whenever any constraint in the batch opened the
+ * normalisation gate, and stayed a correct **0.04** when none did. A hundredfold
+ * understatement decided by an UNRELATED constraint's value. Fixing the
+ * gate-OPEN arm (rather than dragging the gate-closed arm to match it) is what
+ * makes the two agree on the value the producer meant.
+ *
+ * ⚠ AN EARLIER REVISION OF THIS ROW MOVED THE OTHER WAY — it read `'%'` as
+ * always-percentage-points and normalised the sub-1 cell to `v/100`. That was
+ * refuted at the producer's bytes: it would have turned `0.04` into `0.0004` on
+ * the primary draft path, i.e. it would have INTRODUCED the defect above on the
+ * one path that was still correct. The lesson is recorded here because the
+ * mistake is re-makeable from PLoT's side alone: the contract line
+ * ("Threshold in the user's units; PLoT normalises downstream") does NOT settle
+ * which scale `'%'` denotes, and a full mutant kill-rate against that wrong
+ * oracle proves only sensitivity, never correctness.
+ *
+ * ⚠ KNOWN-AMBIGUOUS, deliberately left as CEE has it: a ratio that legitimately
+ * exceeds 100% (NRR 110%) is emitted by the draft prompt as `value: 1.10,
+ * unit: "%"` on a NODE, while CEE's CONSTRAINT rule reads `>= 1` as percentage
+ * points. For a constraint that lands at `1.10` the two readings differ. This
+ * function does NOT resolve it — it reproduces CEE's stated constraint rule
+ * exactly, so PLoT is never the service that invented a third convention.
+ */
+function percentRangeForValue(value: number): NormalisationRange {
+  return Math.abs(value) < 1
+    ? { min: 0, max: 1, source: 'unit_percent' }
+    : { min: 0, max: 100, source: 'unit_percent' };
+}
+
+/**
+ * ROADMAP 2.957 — true when a `'%'` constraint exists whose value the percent
+ * rung WOULD rescale (i.e. percentage-point form, `|value| >= 1`).
+ *
+ * WHY THIS EXISTS, and it is a narrow, derived thing. `constraintsNeedNormalisation`
+ * opens the gate on `value < 0 || value > 1`. `value === 1` satisfies neither,
+ * so a lone `{unit:'%', value:1}` — ONE PERCENT by CEE's `>= 1` rule — was
+ * forwarded raw as `1.0` (a hundred percent) when alone, while resolving to
+ * `0.01` the moment any batch-mate opened the gate. Every other percentage-point
+ * value already opens the gate by magnitude, so this closes the single
+ * remaining cell where the answer still depended on a DIFFERENT constraint.
+ *
+ * Stated as `|value| >= 1` rather than `value === 1` on purpose: it is the
+ * genuine precondition ("the rung would rescale this"), so it stays correct if
+ * the gate's own bounds ever move, instead of encoding today's off-by-one.
+ */
+export function constraintsHavePercentPointValue(
+  constraints: GoalConstraint[],
+  unitsByConstraintId: Map<string, string> | undefined,
+): boolean {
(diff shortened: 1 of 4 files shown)


Talchain:
## Post-open evidence (measured after the PR was raised)

### CI at `fbd6d0f` — concluded, 10/11 green

`gates (macos/ubuntu/windows)` · `test` · `verify` · `guard` · `safety` ×2 · `smoke` · `update_release_draft` — **all success**.
`audit` — **failure**, and it is the standing production-dependency red.

**Empty-delta proof, not a register citation.** Read at my own head (job `93152518773`):

> `fast-uri vulnerable to host confusion via backslash authority introducer — GHSA-7p8r-x3mc-p8w7`
> `4 high severity vulnerabilities`

Then measured directly rather than argued: `npm audit --omit=dev --audit-level=high` run at this branch's HEAD and at pristine `519d3111`, and the two reports are **byte-identical** (`diff` reports no difference). This branch changes **no** `package.json`, `package-lock.json` or `vendor/` file, so the audit surface is untouched by construction as well as by measurement.

### ⚠ Honest gap: no PRE-EXISTING test catches mutant M2

The brief for this row asked that the "break-it-for-a-different-unit" mutant either be caught by existing coverage **or** that the gap be stated plainly. It is the latter, and here is the measurement.

Both runs use the repo's named gate (`npm test`) in an isolated worktree, with **my two new test files deleted** and `constraint-intervention-range-unify.test.ts` reverted to its `519d3111` form — i.e. only coverage that existed before this PR:

| Run | Result | Failing test |
|---|---|---|
| **(a) control** — no mutation | `1 failed \| 7117 passed \| 28 skipped (7157)` | the F1 `'%'` pin (expected — the one cell this PR moves) |
| **(b) M2 applied** — forward-raw rung also exempts `'count'` | `1 failed \| 7117 passed \| 28 skipped (7157)` | **the same test, byte-identical result** |

**Delta from M2 against pre-existing coverage: zero.** Breaking the forward-raw rung for a non-percent unit would have shipped silently before this PR. The only guard that catches it is the one added here — `CONTROL: a NON-percent sub-1 row on a scale-less node is still forwarded RAW, with no diagnostic`.

That is a *pre-existing* coverage gap this PR narrows rather than one it introduces, and it is stated so it is not mistaken for a clean bill.

### Corroboration of the blast radius

The same pair independently re-confirms the claim in the PR body: across all **7,157** pre-existing tests, the *only* one this change moves is that single F1 `'%'` limb — the one amended here, with its `goal_threshold_cap` sibling deliberately untouched.

Talchain:
## CI concluded at `f9aa7657` — 11/12 green

`gates (macos/ubuntu/windows)` · `test` · `verify` · `guard` · `safety` ×2 · `smoke` · `update_release_draft` ×2 — **all success**.
`audit` — **failure**, the standing production-dependency red.

**Empty-delta proof, measured rather than cited.** Read at my own head:

> `fast-uri vulnerable to host confusion via backslash authority introducer — GHSA-7p8r-x3mc-p8w7`
> `4 high severity vulnerabilities`

Then `npm audit --omit=dev --audit-level=high` run at this branch's HEAD and at its base `747a3aa4`: the two reports are **byte-identical** (`diff` reports no difference). This branch changes no `package.json`, `package-lock.json` or `vendor/` file, so the audit surface is untouched by construction as well as by measurement.

## Verification that the reversal actually reversed

The previous revision's core assertion was `{unit:'%', value:0.9} → 0.009`. Mutant **M1** in the current kit re-applies exactly that direction (sub-1 `'%'` never treated as fractional) and is now killed by **6** assertions across both the module suite and the ISL wire — i.e. the tests that would have shipped the old behaviour now actively refuse it.

The F1 pin (`'%' 0.5 → 0.5`) is restored by *not appearing in this diff at all*: `git diff --name-only origin/staging..HEAD` returns only the two `src/` files and the two new test files.

PR #326, opened 2026-08-13
feat(auth): dual-acceptance bearer (ACTIVE + STAGED) for zero-downtime rotation

Talchain:
**DRAFT — do not merge.** An integration freeze is live; opening for review only.

## Why

`authGuard` accepts exactly **one** token, so rotating it forces every caller to change in the same instant the server's value does: CEE's Render env, the UI's Netlify env, and the **`PLOT_AUTH_TOKEN` GitHub Actions secret** used by `staging-smoke.yml` and `load-probe-nightly.yml`. No ordering of those edits avoids a window where something is sending a value the server no longer accepts.

And because `staging-smoke.yml` runs on **every push to staging** and is itself one of those callers, a blind cutover disables the check that would report the breakage it just caused. **The alarm dies with the thing it monitors.**

## The deliverable is the metric, not the second token

Accepting two secrets makes rotation *possible*. `plot_engine_auth_token_match_total{used="active"|"staged"}` is what makes *finishing* it safe — without evidence that nothing matches ACTIVE any more, deleting the old secret is a guess, which is the same failure mode as invalidating blind, only deferred.

Rotation becomes: set STAGED → update callers → watch `used="active"` stop climbing → promote and clear STAGED.

## ⚠ Primary correctness requirement: no hoisted length check, no `||`

The old guard length-checked **once**, against the single expected token, because `timingSafeEqual` throws on unequal lengths — a precondition, not a security check. Extending that naively to two candidates is wrong twice:

1. **Functionally** — a staged token of a different length is rejected before it is considered, so rotation fails for the exact case it exists to serve. A genuinely new secret will not share the old one's length.
2. **As a disclosure** — the work performed would vary with *which* candidate matched on length, making the guard's behaviour a function of a property of the secrets. **A rotation mechanism that leaks which of two secrets you hold is worse than the manual cutover it replaces.**

Each candidate is length-checked and compared **independently**, both evaluated unconditionally before the outcome is chosen.

**The pin is discriminating by construction:** the staged fixture is deliberately a **different length** from the active one. An equal-length fixture would pass against the old shape and prove nothing. RED-first at pristine: 5 failures, including `accepts a STAGED token whose length DIFFERS from ACTIVE` (403 vs 200).

## Copied, not invented

Modelled on the rotation idiom this service already runs for a different secret (`verifyPrincipalSignature` / `incPrincipalSecretFallback`, P0-2) — same shape, same label convention, same test-reset helper — rather than a second mechanism under a different name.

## No secret is ever emitted

The counter records **which** candidate matched, never what it was. `/health` exposes `auth_secrets: {active, staged}` **presence booleans only** — no value, no length, no prefix. Both pinned, and both pins proven by mutants that leak.

## Mutation kit

**8/9 bitten first pass. The survivor was my own defective mutant** (`counts[kind] += 0` leaks nothing, so surviving was correct) — replaced with two genuine leak mutants.

⚠ **One of them exposed a real gap: my "NEVER emits a token value" test was VACUOUS.** `renderAuthTokenMatch()` returns `''` until something matches and `beforeEach` resets the counters, so it asserted *"'' does not contain the token"* — true of every implementation, including one rendering the secret into the exposition. Fixed to drive traffic first, assert non-empty, then assert absence. Both leak mutants now RED. Controls exactly 0 before and after; isolation proven by writing a sentinel; restore from a pristine archive.

## Gate at this tip

- `npm test` (named gate): **exit 0**
- pre-push validation: **6/6 PASS — 7202 passed | 28 skipped**
- typecheck clean · lint clean (`--max-warnings 0`)
- new spec: **17/17**, asserted by name

**Known pre-existing red, not caused here:** `tests/auth.minimal.test.ts` fails identically at **pristine `d39b4fa`** with the same env — verified with a control run in an isolated worktree (same `waitFor` signature, same 2-skipped outcome).

## Scope discipline

Deliberately did **not** touch `/v1`-vs-`/v2` gating or `AUTH_V2_ENABLED`. That was established as a scope toggle rather than a second auth scheme — a fact worth having, not an invitation to tidy.

Also corrects a comment this change made stale: it claimed `PLOT_AUTH_TOKEN` was the *"single source of truth"* with *"no two-var mirror"*, which the code beneath it now contradicts. The single source of truth is the **resolver**, not the number of accepted secrets.

Backwards compatible: with no STAGED set the guard behaves exactly as before; fail-closed when neither is configured.

⚠ **`staging` on this repo is UNPROTECTED** (derived: `branches/staging/protection` → 404), so there is no required status check — the push is the gate. Flagging so the merge decision is made knowingly.

### DIFF
--- a/src/config/auth-token.ts
+++ b/src/config/auth-token.ts
@@ -22,5 +22,38 @@
  * AUTH_ENABLED so it is inert while auth is off.
  */
 export function getExpectedAuthToken(): string {
-  return String(process.env.PLOT_AUTH_TOKEN || process.env.AUTH_TOKEN || '').trim();
+  return String(
+    process.env.PLOT_AUTH_TOKEN_ACTIVE || process.env.PLOT_AUTH_TOKEN || process.env.AUTH_TOKEN || '',
+  ).trim();
+}
+
+/**
+ * The OPTIONAL second accepted secret, for zero-downtime rotation.
+ *
+ * ── WHY A SECOND ACCEPTED VALUE ────────────────────────────────────────────
+ * `getExpectedAuthToken` resolves ONE value, and `authGuard` accepts only that,
+ * so every caller has to change in the same instant the server's value changes:
+ * CEE's Render env, the UI's Netlify env, and the `PLOT_AUTH_TOKEN` GitHub Actions
+ * secret used by `staging-smoke.yml` (which runs on every push to staging) and
+ * `load-probe-nightly.yml`. No ordering of those edits avoids a window in which
+ * something is sending a value the server no longer accepts — and because the
+ * smoke workflow is one of those callers, a blind cutover disables the check that
+ * would report the breakage it caused.
+ *
+ * With STAGED set, rotation becomes: set STAGED to the new value → update every
+ * caller → watch `plot_engine_auth_token_match_total{used="active"}` stop climbing
+ * → promote STAGED into the active name and clear STAGED. Nothing is ever sending
+ * a value the server will not accept.
+ *
+ * The counter is what makes the final step safe: without evidence that nothing
+ * matches ACTIVE any more, deleting it is a guess.
+ *
+ * Same shape as `PRINCIPAL_HMAC_SECRET_STAGED` (P0-2, `verifyPrincipalSignature`),
+ * deliberately — one rotation idiom in this service, not two under different names.
+ *
+ * Empty/unset is the normal state and means "no rotation in progress": the guard
+ * then behaves exactly as before.
+ */
+export function getStagedAuthToken(): string {
+  return String(process.env.PLOT_AUTH_TOKEN_STAGED || process.env.AUTH_TOKEN_STAGED || '').trim();
 }
--- a/src/middleware/auth-guard.ts
+++ b/src/middleware/auth-guard.ts
@@ -9,7 +9,8 @@ import type { FastifyRequest, FastifyReply } from 'fastify';
 import { timingSafeEqual } from 'crypto';
 import { replyWithAppError } from '../errors.js';
 import { isDemoMode } from './demo-mode.js';
-import { getExpectedAuthToken } from '../config/auth-token.js';
+import { getExpectedAuthToken, getStagedAuthToken } from '../config/auth-token.js';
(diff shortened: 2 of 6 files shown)


Talchain:
**Reconciliation note (17 Aug) — SCHEDULE, and this is now the enabler for an outstanding P0 action, not a nice-to-have.**

**Your hold reason has expired.** The body says *"DRAFT — do not merge. An integration freeze is live."* That freeze ended; 20+ PRs have merged across UI and CEE since (UI through #744, CEE through #1001, both deployed). The `do not merge` line is the only thing still holding this, and it is stale.

**Why it now matters more.** UI #682 was closed today as superseded — the PLoT bearer is out of the client bundle and the credential is injected server-side at the Netlify edge (verified live: `/bff/engine/v1/limits` → `200` + `x-plot-proxy-credential: injected`, non-allowlisted paths 404 at the edge). But removing the credential from the bundle **does not un-publish** what every visitor already fetched, so **rotation is still outstanding and is Paul's alone.** `SECURITY-PLOT-BEARER-ROTATION.md` §3 currently asks Paul to choose between *rotate-first* (staging 401s until the deploy lands) and *deploy-first* (the exposed value stays live longer). **This PR removes that choice** — set STAGED, update callers, watch `used="active"` stop climbing, promote and clear. That is precisely the sequencing the runbook cannot offer today, including for the `PLOT_AUTH_TOKEN` GitHub Actions secret that `staging-smoke.yml` uses, which is the caller whose breakage would silence the check that reports it.

**It has not gone stale in the code.** Measured against PLoT staging tip `e0394b869`: `src/config/auth-token.ts`, `src/middleware/auth-guard.ts`, `src/routes/health.ts` and `src/plugins/metrics.ts` are **all present and unmoved** — only **2 commits** have landed on staging since this opened (13 Aug 19:39Z) and **none touches those four files**. Dual acceptance is genuinely absent at the tip: `authTokenMetrics` = 0 files, `auth_secrets` = 0, `PLOT_AUTH_TOKEN_STAGED` = 0. So the diff should still apply close to cleanly; `UNSTABLE` here is check state, not conflict.

Needs: the independent adversarial review this never got, then merge + deploy — **before** Paul rotates, so he inherits the zero-downtime path rather than the dilemma. Rotation itself remains his action; no session should execute it.

PR #327, opened 2026-08-13
fix(trust): drop the placeholder term from published confidence and renormalise (fabrication)

Talchain:
## What & why

The published confidence badge carried a factor its own code marked `// placeholder`. It resolved to a **fixed 0.5 for every production call** (all three callers passed `calibrated: false`, so the term never varied with any input) yet held **15% of the 1000-unit weight budget**. A confidence number the product presents as computed was, by weight, a fabricated constant.

**Scientific integrity — no fabricated confidence.** This drops the placeholder and renormalises the three genuinely-computed factors so the score stays a well-formed value in [0, 1] traceable end-to-end to inputs.

Disposition item 11(b); confirmed by the 13 Aug simplify review.

## Confirmation + the 15% arithmetic (verified at the bytes, tip `d39b4fa`)

`src/trust/confidence.ts` `calculateConfidence()` is the canonical assembler. Integer math, weights sum to 1000:

| Factor | Weight | Source |
|---|---|---|
| identifiability | 350 | computed (`identifiable ? 1.0 : 0.3`) |
| linearity | 250 | computed (`in_linear_range ? 1.0 : 0.5`) |
| k_coverage | 250 | computed (k buckets) |
| **calibration** | **150** | **placeholder — `calibrated ? 1.0 : 0.5`, and `calibrated` is hardcoded `false` at every caller → always 0.5** |

- `calib_raw = round(0.5 × 1000) = 500`; contribution to `overall_raw` = `round(500 × 150 / 1000) = 75/1000`.
- **The placeholder is 15.0% of the weight budget (150/1000)** and adds a **constant +0.075** to every published score in production. As a share of a typical score it runs ~8% (all-strong: 75/925 = 8.1%) up to ~20% (weakest scores). The precise, defensible reading of "~15%" is the **weight allocation**: 15% of the confidence formula is a placeholder constant that never varies with any input.

`calibrated: true` is never passed by any production caller — confirmed at `src/routes/v1/run.ts`, `src/routes/v1/self-check.ts`, `src/inference/model_of_inference.ts` (uses the `false` default), `src/fixtures/demo-payloads.ts`.

## Reader manifest (is it user-reachable?)

Producer → transport → surface, within PLoT (tip `d39b4fa`):

- **Producer:** `calculateConfidence()` — `src/trust/confidence.ts:18`.
- **`/v1/run`:** `src/routes/v1/run.ts:821` builds the badge; placed in the response `base.confidence` at `run.ts:1272`. `confidence.level` also feeds `insights` (`run.ts:1228`).
- **`/v1/self-check`:** `src/routes/v1/self-check.ts:79` → self-check response.
- **`model_of_inference`:** `src/inference/model_of_inference.ts:183`; `confidence.score` feeds `meta_reasoning.quality_assessment.overall_score` at **40% weight** (`model_of_inference.ts:218`) and `confidence_level` (`:232`); `meta_reasoning` surfaces in the `/v1/run` response (`run.ts:1310`).

So the placeholder was **user-reachable twice over**: directly as `confidence.score`/`.level`, and indirectly inside `overall_score`. **Reachable, not dark.**

No downstream code reads `factors.calibration` for any decision (only the producer, the type, and the separate SCM fallback set it). The response JSON schema (`src/schemas/response.ts:7`) is `additionalProperties: true` and does not require `factors`, so dropping the field is wire-safe.

## The fix (one authority)

Fixed only in the canonical assembler `src/trust/confidence.ts`:

- Removed the calibration factor, its weight, its output field, and the now-dead `calibrated` input param.
- **Renormalised the three genuine factors proportionally to sum 1000: 412 : 294 : 294** (`350 × 1000/850 = 412`, `250 × 1000/850 = 294`). Relative importance of the real factors is preserved; the constant floor is gone.
- Dead `calibrated: false` arg removed from the three callers (`run.ts`, `self-check.ts`, `demo-payloads.ts`).
- `ConfidenceBadge.factors.calibration` made optional in `src/trust/types.ts` (the separate SCM-meta fallback at `run.ts:859` still sets its own real `sign_stability` value — left untouched; see Not addressed).

No second normalisation path was added.

## Before / after (calibrated=false, the production value)

Score shifts by −0.02 … +0.07: strong cases rise (the 0.5 placeholder was suppressing them), weak cases fall slightly.

| ident | linear | k | before | after | Δ | level |
|---|---|---|---|---|---|---|
| ✓ | ✓ | ≥1000 | 0.93 | **1.00** | +0.07 | HIGH → HIGH |
| ✓ | ✓ | 500–999 | 0.85 | 0.91 | +0.06 | HIGH → HIGH |
| ✓ | ✓ | <500 | 0.75 | 0.79 | +0.04 | HIGH → HIGH |
| ✓ | ✗ | ≥1000 | 0.80 | 0.85 | +0.05 | HIGH → HIGH |
| ✓ | ✗ | 500–999 | 0.73 | 0.77 | +0.04 | **MEDIUM → HIGH** |
| ✓ | ✗ | <500 | 0.63 | 0.65 | +0.02 | MEDIUM → MEDIUM |
| ✗ | ✓ | ≥1000 | 0.68 | 0.71 | +0.03 | MEDIUM → MEDIUM |
| ✗ | ✓ | 500–999 | 0.61 | 0.62 | +0.01 | MEDIUM → MEDIUM |
| ✗ | ✓ | <500 | 0.51 | 0.51 | 0.00 | MEDIUM → MEDIUM |
| ✗ | ✗ | ≥1000 | 0.56 | 0.57 | +0.01 | MEDIUM → MEDIUM |
| ✗ | ✗ | 500–999 | 0.48 | 0.48 | 0.00 | LOW → LOW |
| ✗ | ✗ | <500 | 0.38 | 0.36 | −0.02 | LOW → LOW |

**Material change disclosed:** exactly **one** production-space level flip — `identifiable=true, outside-linear-range, k=500–999` moves **MEDIUM → HIGH** (0.73→0.77). This is the direct consequence of renormalisation: identifiability's weight rises 35%→41.2%, so a case with strong identifiability legitimately scores higher once the placeholder no longer suppresses it. Disclosed, not hidden; the previously-affected existing test (`trust-signals` "returns MEDIUM") was updated to a genuinely-MEDIUM input with an in-file comment explaining the boundary moved.

## RED-first (named signatures)

New spec `tests/confidence.defabrication.test.ts` — corpus computed by hand (external oracle), covering all-strong / all-weak / mixed / boundary. Against **pristine** source: **10 of 11 assertions FAIL** (the 1 pass is the documented non-discriminating case where before==after score). Discriminating assertions:

- `'calibration' in out.factors` → `false` — REDs at pristine (placeholder emits it).
- all-strong `score === 1.00` — REDs at pristine (capped at 0.93 by the 0.5 term).
- `weakest score === 0.36` (no residual constant floor) — REDs at pristine (0.38 with the +75 addend).

After the fix: **34/34 green** across the new spec + the three updated specs.

## Mutation table (throwaway worktree outside repo root; isolation proven by sentinel write + APFS inode compare; applied-check scoped to `src/`, control exactly 0; restores HEAD-relative)

| # | Mutation | Result |
|---|---|---|
| M1 | ident weight 412 → 350 | **KILLED** (10 failed) |
| M2 | re-add +75 calibration constant addend | **KILLED** (11 failed) |
| M3 | linearity weight 294 → 250 | **KILLED** (10 failed) |
| M4 | re-emit `factors.calibration = 0.5` | **KILLED** (1 failed — precise) |
| M5 | ident weight 412 → 402 (breaks sum=1000) | **KILLED** (7 failed) |

Control: pristine worktree 0 changed files under `src/`, 34/34 green, before and after the run.

## Gate results

- `npm test` (named gate: `node tools/run-all-tests.js` — build + full vitest suite against test server): **exit 0** (run twice).
- `tsc -p tsconfig.json --noEmit`: **exit 0**.
- Full `scripts/pre-push-validate.sh` (7 checks): **exit 0**. Verbatim fast checks: [4/7] 0 stale tracked `.js`; [5/7] `POLICY OK`; [6/7] spectral `0 errors` (5 pre-existing warnings).

## Not addressed (out of scope; candidate for a separate row)

`src/routes/v1/run.ts:859-868` is a **separate constant-confidence path**: when `inferenceResult.meta.unique_graphs` is present it *overrides* the whole badge with a hardcoded `score: 0.6, level: 'MEDIUM'` and `calibration: sign_stability || 0.5` — bypassing `calculateConfidence` entirely. That is a distinct fabrication with its own reachability question and was left untouched to keep this change to one authority. Flagging for a separate assessment.

## Reachability boundary I could not reach from this repo

Within PLoT the badge is definitively produced and placed in the `/v1/run` and `/v1/self-check` responses. Which PLoT route CEE's analyse path actually invokes, and whether the UI renders `confidence.score`/`.level` on the **external PoC surface**, is beyond this repo (v2/run uses a *different* confidence mechanism — `confidence-tier`/`confidence-basis` — not this badge). Not verified here; establishing it needs a live CEE→PLoT capture.

⛔ **DO NOT MERGE** — the deployed quartet is held stable for external review; the orchestrator merges.

### DIFF
--- a/src/fixtures/demo-payloads.ts
+++ b/src/fixtures/demo-payloads.ts
@@ -59,7 +59,6 @@ export function getDemoRunResponse(seed: number = 42): any {
     identifiable: true,
     in_linear_range: true,
     k_samples: 1000,
-    calibrated: false,
   });
 
   const linearity_warning = checkLinearity({
--- a/src/routes/v1/run.ts
+++ b/src/routes/v1/run.ts
@@ -823,7 +823,6 @@ export async function registerRunRoute(app: FastifyInstance) {
         identifiable: identifiability.identifiable,
         in_linear_range: !linearity_warning.outside_range,
         k_samples: K_evaluated ?? budget.k,
-        calibrated: false,
       });
 
       // Threshold crossings with actual p50
--- a/src/routes/v1/self-check.ts
+++ b/src/routes/v1/self-check.ts
@@ -81,7 +81,6 @@ export async function registerSelfCheckRoute(app: FastifyInstance) {
       identifiable: identifiability.identifiable,
       in_linear_range: !linearity_warning.outside_range,
       k_samples: budget.k,
-      calibrated: false,
     });
 
     // Threshold crossings
--- a/src/trust/confidence.ts
+++ b/src/trust/confidence.ts
@@ -9,11 +9,18 @@ export interface ConfidenceInputs {
   identifiable: boolean;
   in_linear_range: boolean;
   k_samples?: number;
-  calibrated?: boolean;
 }
 
 /**
- * Calculate confidence score based on graph quality and coverage
+ * Calculate confidence score based on graph quality and coverage.
+ *
+ * Every term below is a computed function of the inputs. The former
+ * "calibration" factor was a placeholder that resolved to a fixed 0.5 for
+ * every production call (all callers passed `calibrated: false`, so it never
+ * varied) yet carried 15% of the weight budget — a constant standing in for a
+ * computation. It has been removed and its weight redistributed across the
+ * three genuinely-computed factors, proportionally, so the score remains a
+ * well-formed value in [0, 1] traceable end-to-end to inputs.
  */
 export function calculateConfidence(inputs: ConfidenceInputs): ConfidenceBadge {
   const {
@@ -21,7 +28,6 @@ export function calculateConfidence(inputs: ConfidenceInputs): ConfidenceBadge {
     identifiable,
     in_linear_range,
     k_samples = 1000,
-    calibrated = false,
   } = inputs;
 
   const node_count = graph.nodes.length;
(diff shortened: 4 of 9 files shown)


Talchain:
Closing in favour of #328 (merged as a5345a5e), adjudicated by comparison review: #328's threshold rescale preserves all 12 reachable verdicts where this PR's fixed thresholds produced a unit-mismatch flip (MEDIUM→HIGH on a linearity-caveated cell), and its snapshot test fails at this head. This PR's independent value was real and three of its ideas were grafted into #328 before merge: exact per-cell score pins (closing a measured 2,419-reweighting hole), the exact factors key-set assertion, and the dead-calibrated-args tidy. Both lanes independently found the run.ts:856 SCM fabrication (rowed 2.1140).

PR #329, opened 2026-08-17
fix(honesty): flip-aware robustness caveat; no fabricated review defaults (2.1247+2.1248)

Talchain:
## ROADMAP 2.1247 + 2.1248 — analysis-honesty on the wire

Feeds the ratified acceptance target: *never present conflicting claims about whether analysis ran, is fresh or can support a conclusion*. Derived and fixed at base `a5345a5e`.

### 2.1247 — `robustness_caveat` was structurally blind to flip evidence

**Mechanism (confirmed at the bytes at `a5345a5e`):** `src/assembly/decision-brief.ts:695-722` composed the caveat from robustness marginals alone; `BriefAssemblyInput` (`:121-149`) had no `flip_thresholds` member, so the caveat could ship *"small changes to assumptions could change which option leads"* while the same payload's flip evidence attested (`structurally_invariant` / `no_effect_within_bounds`) that no tested factor can flip the leader — the same self-contradiction witnessed live for the display verdict and fixed for its reason string in ROADMAP 2.278.

**Derivation refinement (2.1247's second clause):** the display-verdict side has **no remaining blindness** — `robustness-display-verdict.ts` already consumes the same run's flip evidence for its reason (2.278) and deliberately, correctly never moves the verdict on it (`:40`, `:82-89`). No change made there; the file's documented behaviour is the ratified pattern this PR extends to the brief.

**Fix — two named claims that cannot contradict (trap-21: never two questions under one name):**
- **Claim 1** (`robustness_caveat.text`) — aggregate stability under the perturbations tested, marginals scope, unchanged basis semantics. On an attested no-flip (`all_no_effect`), the two branches that previously asserted a flip drop the flip language but keep their marginal verdict (mirrors 2.278's correction shape; verdict never moves).
- **Claim 2** (`robustness_caveat.flip_evidence`, additive optional) — `{status, text}`, status **derived via `classifyFlipThresholdsStatus`** (single source of truth, never re-derived — the hand-maintained-mirror class). Present only when probes support a claim (`computed` / `all_no_effect` / `partial_no_effect`); absent on `unavailable`/`unresolved` — an unfinished probe attests nothing. Wording claim-safe (no numbers), scoped to the probed set per the 2.292 scoping ruling ("the factors we could test", never a universal).
- `run.ts` threads the **same `flipThresholds` variable** the response publishes and the display verdict consumes — one array, so the caveat can never cite a different run's evidence.

Consumer safety: `decision_brief` is an open passthrough in the shared contract (`@talchain/schemas` 0.40.0: `z.object({}).passthrough().nullable().optional()`); `flip_evidence` is additive-optional; with flip evidence absent the caveat is **byte-identical** to its pre-change shape (pinned by test). Golden fixtures unaffected (all four carry recognised levels; the live golden's flip status is `unavailable`).

### 2.1248 — fabricated defaults on the review path

Consumer contract derived before changing anything: CEE's `DecisionReviewInputSchema` (read-only at CEE staging tip `c5e24307`, `src/routes/assist.v1.decision-review.ts:150-224`) admits `isl_results.robustness` as `z.record(z.unknown()).optional()` — absent keys are admitted — and requires `winner` **non-null** `{id, label, win_probability}`. CEE serialises `isl_results` and `winner` **verbatim into the reviewing model's prompt** (`buildDecisionReviewUserMessage`), so every fabricated default below reached the model as a measured figure. The vendored `@talchain/schemas` 0.40.0 pin carries no DecisionReviewRequest schema (checked at the tarball) — CEE's Zod is the authority.

**Absent-propagation table:**

| Field | Old default | New behaviour | Consumer effect |
|---|---|---|---|
| `decision_brief.robustness` (level absent/unrecognised) | `'moderate'` — fabricated middle value | `'not_assessed'` | Same honest token the sibling `robustness.display_verdict` already ships to the same consumers for the same runs; shared contract is open passthrough. `'medium'` now an explicit case (ISL V2 vocabulary — previously mapped correctly only *by accident of the fabricating default*) |
| `isl_results.robustness.is_robust` | `?? false` — "not robust" claimed for unassessed runs | key omitted | CEE schema admits; prompt JSON block simply lacks the key; a **measured** `false` preserved verbatim |
| `isl_results.robustness.recommendation_stability` | `?? 0` — "0% stability" | key omitted | As above; a measured 0 preserved |
| `isl_results.robustness.flip_risk_category` | `?? 'unknown'` — presented as ISL's categorisation | key omitted | As above |
| `winner` (no analysed options) | `{id:'', label:'', win_probability: 0}` sent to CEE | request not built (`buildDecisionReviewRequest` returns `null`); orchestrator skips with named reason `NO_ANALYSED_OPTIONS` | CEE requires non-null winner, so honest absence is not representable on this wire — visible skip instead of confident fabrication. `review_skip_reason` is a PLoT-owned union (`engine-v3.ts:1615`); additive value, not pinned in the schemas tarball (target-zero + contrast-positive sweep) |
| corrector source `robustness.recommendation_stability` (`decision-review-orchestrator.ts:459`) | `?? 0` — authoritative correction source | emitted only when measured | number-corrector can no longer "correct" a model figure **to** a fabricated 0, or bless a written "0%" as grounded |
| Tier-4 grounding allowlist (`m1-review-validator.ts:912`) | fabricated 0 pushed unconditionally | pushed only when a number was measured | a review citing "0" stability for an unassessed run now trips `UNGROUNDED_NUMBER` (warning-grade) instead of passing as grounded |

**Refinement to the row's wording:** "winner/runner-up fall back to `{id:'',…}`" — at `a5345a5e` only **`winner`** fabricated (`:182`); `runner_up` already fell back to honest `null` (`:192`). Fixed the real half.

**Same-family fixes included with their own file:line** (named mechanisms' twins on the identical review path, not scope expansion): `buildIslResultsForCorrection` `?? 0` and the validator's unconditional allowlist push — both launder the same fabricated stability into the same review.

**Deliberately NOT changed (boundary):** `extractOptionComparison`'s `win_probability ?? 0` (`decision-review-request.ts:273`) and `buildIslResultsForCorrection`'s option `win_probability ?? 0`. CEE's schema requires `winner.win_probability: z.number()` (non-optional), and win_probability drives winner selection, margin, and ranking semantics — honest absence is not representable for it on this wire without a CEE-side schema change. Smallest enabling change if wanted: CEE makes `winner.win_probability` optional (or admits a `win_probability_basis` marker); until then rows with absent win probabilities keep today's behaviour. Reported, not forced.

### Evidence

- **RED-first at pristine `a5345a5e`:** 15 named failures across 4 files (contradiction kill, domain-wide consistency matrix, absent-key honesty, null-winner, corrector/allowlist, `not_assessed` mapping, orchestrator skip — the skip test additionally proved the fabricated request was previously **sent** to CEE). All GREEN post-fix (97/97 in the four files, same collection count as RED).
- **Spec-level invariant (trap 13d):** the domain matrix runs every `is_robust × level × flip-class` combination (126 cells) and asserts claims cannot contradict + `flip_evidence.status` equals the shared classifier's output — written against the consumer spec, not the witnessed failure.
- **Mutant kit — 12/12 BITTEN** (throwaway worktree outside the repo root, isolation proven by sentinel WRITE test + inode comparison before any mutant ran, HEAD-relative restores, applied-check scoped to `src/` asserting exactly the intended file, clean-tree assertion before/after every mutant, trailing zero-diff GREEN control 97/97):
  - M1 `?? 'moderate'` restored → 2 tests RED · M2 `?? false` → 2 RED · M3 `?? 0` (extract) → 2 RED · M4 `?? 'unknown'` → 2 RED · M5 empty-winner fallback restored → 3 RED (incl. proof the request would again be SENT to CEE) · M6 `?? 0` (corrector source) → 1 RED · M8 run.ts de-threads `flip_thresholds` → structural pin RED · M9 caveat ignores flip evidence → 3 RED · M10 flip claim emitted for unresolved → 2 RED · M11 local status re-derivation misfiling unresolved as attested (mirror-drift simulation) → 2 RED · M12 all_no_effect wording swapped for computed's (scope lie) → 2 RED.
  - **M7 (unconditional allowlist push) initially SURVIVED and was NOT waved through as equivalent** (trap 13c: equivalence must be demonstrated): with `extractRobustness` fixed, the restored push inserts `undefined` rather than the old fabricated 0, which a `not.toContain(0)` guard cannot see — a guard watching one door. The test now also pins the SPEC (every allowlisted entry is a finite number, non-empty), and M7 REDs against the amended tip.
- **Gates at tip (derived from `.github/workflows/ci.yml`: stale-js → lint → validate:contracts → build → test):** lint clean, contracts validated, build clean, typecheck clean.
- **Full suite with delta attribution:** pristine `a5345a5e` control (fresh worktree, machine quiet): **7202 passed / 0 failed / 28 skipped**. Lane commit, machine quiet: **7225 passed / 0 failed / 28 skipped** — +23 tests, exactly this PR's additions; **empty failure delta**. (An earlier run of the lane commit showed 1 perf-gate red + 8 server-boot `waitFor` file timeouts with 78 hook-skipped tests; it overlapped a concurrent `npm ci` in another worktree, and the quiet re-run proved all of it environmental rather than asserting it.) Full-suite numbers were measured at `6007b11`; the amend to `2b12fa1` is test-only in one file, re-run green at tip (10/10 + 97/97 four-file control); CI on this PR is the authoritative at-tip gate.
- **Collateral:** `tests/fragile-edges.cee-order.test.ts` fixture carried no analysed options (its fabricated empty winner rode along unnoticed); now carries minimal real options with a non-null constructibility assertion. One historic-corpus check: no capture corpora touched (trap 14b).

### DIFF
--- a/src/assembly/decision-brief.ts
+++ b/src/assembly/decision-brief.ts
@@ -15,7 +15,7 @@
  * top_drivers:       factor_sensitivity[] top 5 by abs(elasticity)
  * key_assumptions:   m1_coaching.evidence_gaps[].factor_label → []
  * what_would_change: robustness.fragile_edges[].from_label/to_label → factor_sensitivity[].factor_label (sorted by |elasticity|)
- * robustness:        robustness.level → 'moderate' (default)
+ * robustness:        robustness.level → 'not_assessed' when absent/unrecognised (2.1248 — never a fabricated 'moderate')
  * warnings:          critiques (severity>=warning) + m1_coaching.model_critiques → []
  * lineage.response_hash: meta.response_hash (computed before brief assembly)
  * lineage.config_version: SHA-256 hash of n_samples_default + review/facts flags + brief_assembly_version
@@ -43,6 +43,10 @@ import { NEAR_TIE_THRESHOLD } from '../trust/result-coherence.js';
 // A1b: intervention-controlled levers are not independently tunable; exclude them
 // from the |elasticity|-ranked tunability surfaces (top_drivers, what_would_change).
 import { filterInterventionOverrides, interventionOverrideFactorIds, filterLeverSourcedFragileEdges } from '../lib/intervention-override.js';
+// ROADMAP 2.1247: the single source of truth for flip-threshold classification
+// — the caveat's flip claim derives its status here, never re-reads
+// flip_reason strings (the hand-maintained-mirror defect class).
+import { classifyFlipThresholdsStatus } from '../lib/flip-threshold-status.js';
 
 // =============================================================================
 // Constants
@@ -140,6 +144,16 @@ export type BriefAssemblyInput = Pick<RunResponseV3, 'analysis_status' | 'critiq
    * (DEFAULT-coded disclosures). Absent → both surfaces treat as none.
    */
   inference_warnings?: RunResponseV3['inference_warnings'];
+  /**
+   * ROADMAP 2.1247 (additive, optional): the SAME denormalised flip-threshold
+   * array the response publishes at `flip_thresholds` — run.ts passes the one
+   * variable both read, so the caveat and the evidence it cites can never come
+   * from two different runs (the same-run doctrine as `display_verdict_reason`,
+   * ROADMAP 2.278). Feeds `robustness_caveat.flip_evidence` (claim 2) and the
+   * attested-no-flip rewording of claim 1. Absent → classified 'unavailable'
+   * → caveat byte-identical to its pre-2.1247 shape.
+   */
+  flip_thresholds?: RunResponseV3['flip_thresholds'];
   response_hash?: string;
   meta: {
     seed_used: string;
@@ -166,18 +180,27 @@ export type BriefAssemblyInput = Pick<RunResponseV3, 'analysis_status' | 'critiq
  *   - low driver confidence and near-tie status → tone gate `headline`.
  *
  * The semantic mapping is "graph perturbation stability under ISL's
- * robustness analysis" — `high → 'robust'`, `moderate → 'moderate'`,
+ * robustness analysis" — `high → 'robust'`, `medium | moderate → 'moderate'`
+ * (ISL V2 wire vocabulary is 'medium'; 'moderate' tolerated — the same
+ * normalisation `deriveVerdict` applies in robustness-display-verdict.ts),
  * `low | very_low → 'fragile'`. See `src/types/decision-brief.ts` jsdoc
  * for the consumer-facing contract and the readiness-widening follow-up
  * for the longer-term action-readiness surface.
+ *
+ * ROADMAP 2.1248: an ABSENT or unrecognised level maps to `'not_assessed'`
+ * — never to a fabricated `'moderate'`. Before this fix the `default` branch
+ * did double duty: it correctly normalised 'medium' AND silently converted
+ * "nobody measured this" into a confident middle value. 'medium' is now an
+ * explicit case, so the default only ever sees genuinely unassessable input.
  */
-function mapRobustnessLevel(level: string | undefined): 'robust' | 'moderate' | 'fragile' {
+function mapRobustnessLevel(level: string | undefined): 'robust' | 'moderate' | 'fragile' | 'not_assessed' {
   switch (level) {
     case 'high': return 'robust';
+    case 'medium':
     case 'moderate': return 'moderate';
     case 'low':
     case 'very_low': return 'fragile';
-    default: return 'moderate';
+    default: return 'not_assessed';
   }
 }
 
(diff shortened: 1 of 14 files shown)


Talchain:
**CI at head `2b12fa1b` — polled to conclusion (exact-SHA check-runs, 11 total, 0 running):** 10 success (gates ×3 OS, guard, safety ×2, smoke, test, verify, update_release_draft) · 1 failure: `audit`.

**Audit attribution (read at the job log, not from a registry):** `npm audit --omit=dev --audit-level=high` fails on **fast-uri host-confusion advisory GHSA-7p8r-x3mc-p8w7** (4 high). This PR's dependency surface is byte-identical to base `a5345a5e` (`git diff a5345a5e..2b12fa1 -- package.json package-lock.json vendor/` → **0 files**), so the red is advisory-driven, pre-existing, and estate-rowed (ROADMAP 2.385). No base-commit audit check-run exists to witness the same red on staging directly — stated as the limit of this evidence rather than papered over.

Full mutant log, RED-at-pristine listing, and full-suite delta-attribution logs live in the lane workspace; summarised in the PR body. **Do not merge without independent review** (lane instruction — PLoT staging is unprotected; the push is the gate).

